Issue #46, opened 2015-11-19
Add CommandLine parameter which allows to pass Token to login commandw


forki:
We want to automate the release process of the F#, FAKE and Paket plugins. Currently we fail to login since the vsce login is asking for the Personal Access Token. It would be much easier for us if we could give it as optional parameter to vsce login command.

joaomoreno:
:+1:

joaomoreno:
Fixed per #47.

PR #47, opened 2015-11-19
Add Personal Access Token as optional parameter in login command

Krzysztof-Cieslak:
#46. //CC : @joaomoreno @forki

Unfortunately, there are no tests for login command (and it's also bit hard to test it for me by hand), so just hoping it works ;)

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -55,7 +55,8 @@ module.exports = function (argv: string[]): void {
 	program
 		.command('login <publisher>')
 		.description('Add a publisher to the known publishers list')
-		.action(name => catchFatal(loginPublisher(name)));
+        .option('-p, --pat [token]', 'Personal Access Token')
+		.action((name,pat) => catchFatal(loginPublisher({ publisher: name, pat: pat})));
 
 	program
 		.command('logout <publisher>')
--- a/src/store.ts
+++ b/src/store.ts
@@ -25,6 +25,11 @@ export interface IGetOptions {
 	promptIfMissing?: boolean;
 }
 
+export interface ILoginOptions {
+	publisher?: string;
+	pat?: string;
+}
+
 function load(): Promise<IStore> {
 	return readFile(storePath, 'utf8')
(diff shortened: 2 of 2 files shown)


forki:
:heart:

joaomoreno:
Closed and reopened to get the CLA bot working.

forki:
@joaomoreno any technical comments?

joaomoreno:
Absolutely. I will review this now. Build looks good, just need the CLA from you. :+1:

joaomoreno:
Actually, I'm thinking that the better fix for this is just to allow passing in the `pat` as an option to all server related commands, instead of doing it in the login command. That's a better story for automation since it doesn't involve writing to `~/.vsce`.

Let me sketch that up...

forki:
Yes if we could avoid a "login" step altogether that would be even better.

BTW:
https://github.com/ionide/ionide-vscode-paket/blob/master/build.fsx#L139
shows our build script that transpiles a F# addin to javascript and
publishes it on the gallery and github.

joaomoreno:
Very cool.

Closing this PR, let's keep discussing it here though.

I've published `vsce@0.9.0-beta`. You wanna give it a try in your build script? You can get it by running `npm install -g vsce@beta`. Let me know how it goes.

forki:
image

doesn't seem to work with vsce@0.9.0-beta

joaomoreno:
Sure sure? What do you get with `vsce -V`?

forki:
image

forki:
ah. it's still on 0.8.1. 

but the build is installing from beta channel:

image

how can I call the new one?

joaomoreno:
No idea. Forget the beta, I just published 0.9.0, give it a try.

forki:
mhm. it downloads 0.9.0 but still calls 0.8.1 so the error is on my side and I'm just calling wrong version of vcse (seems I have two installed)

forki:
found the correct version. now it works like a charme. thanks a lot

joaomoreno:
Awesome!

Issue #52, opened 2015-12-15
Ignore dependencies of devDependencies


SrTobi:
Are dependencies of devDependencies ignored?

When I use `npm install` and then `vsce ls` a lot of node_modules are still listed. the devDependencies are ignored but their dependencies are not.

Wouldn't it be better to include all production dependencies and their production dependencies and ignore all other modules?

felixfbecker:
Very annoying! [The docs](https://code.visualstudio.com/docs/tools/vscecli) state:
 > Note: Development dependencies listed in devDependencies will be automatically ignored, you don't need to add them to the .vscodeignore file.

But when I check the installed extension:
```
~\.vscode\extensions\felixfbecker.php-debug-1.1.0$ npm ls
php-debug@1.1.0 C:\Users\felix\.vscode\extensions\felixfbecker.php-debug-1.1.0
├── commander@2.3.0 extraneous
├── debug@2.2.0 extraneous
├── diff@1.4.0 extraneous
├── escape-string-regexp@1.0.2 extraneous
├── file-url@1.0.1 extraneous
├── glob@3.2.3 extraneous
├── growl@1.8.1 extraneous
├── iconv-lite@0.4.13
├── jade@0.26.3 extraneous
├── mkdirp@0.5.0 extraneous
├── moment@2.11.1
├── supports-color@1.2.0 extraneous
├── url-relative@1.0.0
├── vscode-debugadapter@1.2.0
├── vscode-debugprotocol@1.2.1
└── xmldom@0.1.20
```
See all the "extraneous" packages installed, those are dev-dependencies-dependencies.

felixfbecker:
Here is a workaround, put this is package.json scripts:
```json
"vscode:prepublish": "npm install && npm run compile && npm prune --production"
```
Before publishing, it will install everything, then compile, and at the end remove all devDependencies. Unfortunately, there is no postpublish hook, so you will have to run `npm install` after publishing again.

SrTobi:
Great idea :+1:

joaomoreno:
That is really cool, @felixfbecker, didn't know about `npm prune`.

heaths:
Not really a fix, though. Perhaps something like this (opened before I realized there was a separate repo for vsce): https://github.com/Microsoft/vscode/issues/2114

felixfbecker:
I dont think reinstalling on publishing is the right solution.
[...]

PR #58, opened 2015-12-18
Improve dependency inclusion

SrTobi:
The current ignore algorithm for devDependencies is pretty simple and does not really work for vscode extensions. That's mainly because productive dependencies of devDependencies are not ignored and vscode has a lot of them. I tried to fix this by using npm itself: using the `ls` command npm builds a dependency structure that can be used to extract all real productive dependencies. In a second step all files in the selected modules are globed and put into a list. 

It works quite well and has the advantage that also modules that are in other modules treated in a correct way. There are still a few issues though:

- The main problem is that until now I have not found a way to silent the npm output
- Because of npm the nodejs tsds where updated... I don't know if thats ok

So this pr is more of a first draft to fix the overall problem.


SrTobi:
I also opened an issue few days ago #52.

joaomoreno:
This is really clever, thanks! We will evaluate it for the next release!

felixfbecker:
I think this is doing things way to complicated. NPM has an option to restrict packages in the list to only include production dependencies. Then all other folders should be excluded.

joaomoreno:
@SrTobi, sorry for taking time to get to this. I think it is better here to use npm as a CLI instead of a dependency, simply because the transitive closure of dependencies of npm is huge.

Also, the definition file for Node upgrade is incorrect, since vsce needs to support 0.12.

Let me know if you want to update your PR. Otherwise, I will pick it up and use the CLI call.

felixfbecker:
@joaomoreno That doesn't make much sense, npm is just an npm module itself. Since it is already installed on the user's system, its dependencies are installed already too. Since vsce is a global package just like npm, all the dependencies are already installed.

When you use the API, you a) spawn an unneeded process and have to do unneeded parsing, which means worse performance and b) you make a "global" dependency on the npm module without specifying it in package.json.

felixfbecker:
@SrTobi did you implement the suggestions I raised about using the production-only parameter instead of custom parsing?

SrTobi:
@fabienroyer nope I haven't touched the code since i opened the pr.

SrTobi:
The idea behind using the npm module is, that the interface of the npm api is less likely to change in comparison to the output of npm (ok, ok it was also faster to implement :smile:). This however was only my feeling and needs further investigation. I also do not like the version problem... this is something we definitely avoid if we use the npm cli.

Let me investigate, if the output of `npm list` has changed since v0.12

felixfbecker:
@SrTobi what do you mean with "the version problem"? And why exactly does vsce need to support an ancient Node version 0.12 when VS Code itself runs on 4.1.1?

SrTobi:
By the "version problem" i mean the problem that we must support npm version 0.12. That might be problematic because if vsce uses npm version v0.12 but the system runs a much higher version of npm, the npm module might not be able to handle the node_module filesystem structure as well as other things.

@joaomoreno might know more about why we have to support version v0.12.

SrTobi:
Ok, I changed the PR. Now it uses an child process and npm without the npm module.

joaomoreno:
Awesome job, @SrTobi, thanks for the time and effort. Just published `vsce@1.1.0` that has these changes.

@felixfbecker I just think there are still a bunch of people out there on Node < 4 installed in their systems, and I would like to keep compatible with those guys until a bit more time passes.

SrTobi:
@joaomoreno have you tested it with older versions of npm?

joaomoreno:
Runs fine with `npm@2.14.16`

v4run:
Hi, I am still facing this issue for an extension I wrote. `vsce package` makes a 5MB file. The only thing that worked for me is

```javascript
"vscode:prepublish": "npm install && npm run compile && npm prune --production"
```
which creates a much smaller file.

I am using `vsce@1.1.0` and ` npm@3.7.1`

PR #60, opened 2016-01-23
Added options to collect call in pack function

hoovercj:
The optional parameter wasn't being passed when calling collect so baseImagesUrl and baseContentUrl weren't being respected.

I also added a test to verify that baseImagesUrl is handled properly by the processor.


joaomoreno:
Thanks! :+1:

PR #61, opened 2016-01-28
added typescript devDependency

SrTobi:
Is it just me or is typescript missing in the devDependencies?
I always have to install it manually.


joaomoreno:
:+1: `npm@3` needs peer dependencies listed explicitly

PR #65, opened 2016-02-19
removed --depth to fix #64

### LINKED ISSUE #64: vsce@1.1.0 publishes incomplete, unusable packages in some cases
`DavidAnson.vscode-markdownlint-0.4.0` was created with `vsce@1.1.0` (on Windows) and does not work in VS Code because dependencies of dependencies were not included in the VSIX.

Specifically, the following directories are missing:
* `DavidAnson.vscode-markdownlint-0.4.0/node_modules/opn/node_modules/pinkie-promise/node_modules/...`
* `DavidAnson.vscode-markdownlint-0.4.0/node_modules/markdownlint/node_modules/markdown-it/node_modules/...`

`DavidAnson.vscode-markdownlint-0.4.1` was created with `vsce@1.0.0` using the same extension sources and includes those dependencies in the VSIX (and therefore works).

Looking at the commit history for vsce between `1.0.0` and `1.1.0` shows significant changes to how `node_modules` directories are handled.

Unless specific to the `markdownlint` extension, this problem seems like it would affect all non-trivial (i.e., have dependencies with dependencies) VS Code extensions published with `vsce@1.1.0`.

If you want to experiment, this is the markdownlint extension repository: https://github.com/DavidAnson/vscode-markdownlint

SrTobi:
We still have some problems with this `npm list`, but this should fix at least the problems in #64.

### DIFF
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -1,7 +1,7 @@
 import * as path from 'path';
 import * as cp from 'child_process';
 
-const cmd = 'npm list --production --parseable --depth';
+const cmd = 'npm list --production --parseable';
 
 export function getDependencies(cwd: string): Promise<string[]> {
 	return new Promise<string[]>((c, e) => {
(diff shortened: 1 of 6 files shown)


SrTobi:
added some tests to verify #64

joaomoreno:
Awesome, thanks for this!

PR #73, opened 2016-02-29
Output warning when using bad npm version

### LINKED ISSUE #72: Dependency listing does not work in npm@3.7.0-3.7.3
As mentioned in https://github.com/Microsoft/vscode-vsce/issues/52#issuecomment-183629911 the inclusion of dependencies of dependencies does not work with some npm versions. I did some testing and found out, that it does not work with `3.7.0` - `3.7.3`. Since `3.7.4` it works again.

Maybe we should output a warning when using a npm version, that does not work.

SrTobi:
To resolve #72.

Maybe the warning should be outputted directly at the start of vsce.

### DIFF
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -1,11 +1,11 @@
 import * as path from 'path';
 import * as cp from 'child_process';
 
-const cmd = 'npm list --production --parseable';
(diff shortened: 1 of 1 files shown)


joaomoreno:
:+1: Thanks!

Issue #78, opened 2016-03-02
Create link command


Krzysztof-Cieslak:
Apm (similar tool for Atom editor) has `link` command which creates symlink between current folder and Atom packages folder - very useful feature for developers building editor plugins. 
![image](https://cloud.githubusercontent.com/assets/5427083/13479140/4ccffbcc-e0d5-11e5-8103-d021778fce49.png)

We could implement same functionality, just for VSCode plugin folder. I should be able to PR it, if this suggestion is accepted.

joaomoreno:
Interesting. cc @bpasero 

We have many targets though, we would likely need an extra optional argument to it indicating which Code installation to use: Stable (default), Insiders, Alpha, OSS, OSS Built

joaomoreno:
@Krzysztof-Cieslak Why don't you submit that PR and I'll make sure it covers all bases as to all installation targets we have?

bpasero:
:+1: 

The only thing that confuses me with link in npm usually is that you need a command to find out what kind of links you have so that you know how to clean up :)

PR #79, opened 2016-03-04
[WIP] Link command implementation

Krzysztof-Cieslak:
As discussed in #78. I only know what are correct paths for `stable` and `insiders` target.

### DIFF
--- a/src/link.ts
+++ b/src/link.ts
@@ -1,0 +1,47 @@
+import * as _read from 'read';
+import * as fs from 'fs';
+import * as path from 'path';
+import * as util from './util';
+
+export function linkPlugin(dir : string, targ : string) : Promise<any> {
+	return new Promise((c, e) => {
+		let packagePath = dir || "."
+		let target = targ || "stable"
+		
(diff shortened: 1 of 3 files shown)


SrTobi:
What happens when I click uninstall in VSCode?

Krzysztof-Cieslak:
It just removes symlink

SrTobi:
:+1: that's cool

stkb:
It's a bug in vscode, but on Windows, uninstalling a symlinked extension deletes the contents of the target directory! (Just comfirmed on 0.10.8 and 0.10.10-insiders)

felixfbecker:
The path for the alpha build is in `~/.vscode-alpha` and for self-built from source `~/.vscode-oss-dev`

joaomoreno:
Sorry for taking so long to pick this up... We definitely want to do this, but it is now dependant on Microsoft/vscode#3884, so let's wait a bit more until that one is completed.

Krzysztof-Cieslak:
No worries at all. Thanks for all crazy good work! :)

Issue #90, opened 2016-05-14
Can no longer publish pre-packaged extension


hoovercj:
Commit  6027e49 broke the ability to publish an already packaged extension. In doing so, it also broke the ability to set `--baseImagesUrl` and `--baseContentUrl` (which was only available in `vsce package`) which I relied on for my publishing process.

PR #91, opened 2016-05-14
Add --packagePath to publish command

### LINKED ISSUE #9: Auto-increment package version?
From @lukehoban:

One small but nice thing that Atom’s publishing does is to auto-increment version number based on a version increment flag and commit/tag/push that in the repo before packaging.

This is definitely more opinionated – but makes it really convenient to re-publish frequently, and avoids accidentally re-publishing without incrementing version number.

Here's a snippet from the APM docs that talks about the behaviour.

```
Usage: apm publish [<newversion> | major | minor | patch | build]
       apm publish --tag <tagname>
       apm publish --rename <new-name>

Publish a new version of the package in the current working directory.

If a new version or version increment is specified, then a new Git tag is
created and the package.json file is updated with that new version before
it is published to the apm registry. The HEAD branch and the new tag are
pushed up to the remote repository automatically using this option.
...
```

Would it makes sense to add something similar for to `vsce`'s publish command?

hoovercj:
Addresses #90 which was a breaking change introduced by 6027e49 to fix #9

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -26,7 +26,8 @@ module.exports = function (argv: string[]): void {
 		.command('publish [<version>]')
 		.description('Publishes an extension')
 		.option('-p, --pat <token>', 'Personal Access Token')
-		.action((version, { pat }) => catchFatal(publish({ pat, version })));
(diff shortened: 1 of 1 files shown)


joaomoreno:
Thanks, sorry for the breakage!

PR #94, opened 2016-06-01
Full parity on keywords with MP

joaomoreno:
From @waderyan, #93

MP folks are running an implementation of this pseudocode on the INT database. Once we confirm the data looks good we'll run it on prod. 

We will want to bring full parity with what happens on update for VSCE. The psuedocode is [here](https://github.com/Microsoft/vscode-marketplace/blob/master/tag-seeding-algorithm.js).

Couple things that have evolved since #83. 
- [x] if we detect `contributes.keybindings` tag with `keybindings`
- [x] if we detect `contributes.debuggers` tag with `debuggers`
- [x] if we detect `contributes.jsonValidation` tag with `json`
- [x] use alias to keword map (see [here](https://github.com/Microsoft/vscode-marketplace/blob/master/tag-seeding-algorithm.js#L7)) on description (see [here](https://github.com/Microsoft/vscode-marketplace/blob/master/tag-seeding-algorithm.js#L103))
- [x] if we detect `contributes.grammars.language`, tag with `{ language }`
- [x] if we detect `contributes.languages` then, tag with `{ language.id }`, ` { language.aliases }`, and create hidden keywords for `{ language.extensions }`. Hidden = `$ext_{ language.extensions }`. See [here](https://github.com/Microsoft/vscode-marketplace/blob/master/tag-seeding-algorithm.js#L66). 

My thought is let's wait until MP engineers run this on the INT environment before we make the changes. 

Small update from MP.

- [x] Change hidden tags to '__ext_`{language.extensions}`'
- [x] Make sure we don't have the "java" "javascript" issue

PR #96, opened 2016-06-15
Issue 93 change hidden tag

waderyan:
@joaomoreno Here are a couple fixes for #94. 

First I changed the hidden tag to be '__' instead of '$'. That was easy. 
The second task I'm having a harder time with and could use your help. An issue we saw with the MP algorithm was javascript extensions being tagged with java. I attempted to fix that by sorounding the regex with '\b' - that solves the "java" issue but introduced a "c++" tagging issue. 

The test is failing now on tagging an extension as "c++". Would love if you could take a look at it. 

Some things I tried. 
- Escaping the regex per instructions [here](http://stackoverflow.com/questions/2593637/how-to-escape-regular-expression-in-javascript) and [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions) - because "+" is a regex character. The function from Mozilla didn't seem to work.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "vsce": "out/vsce"
   },
   "scripts": {
-    "test": "mocha",
+    "test": "gulp compile && mocha",
     "prepublish": "gulp compile && mocha"
   },
   "engines": {
--- a/src/package.ts
+++ b/src/package.ts
@@ -130,7 +130,6 @@ export class TagsProcessor extends BaseProcessor {
 		'react': ['javascript'],
 		'js': ['javsacript'],
 		'node': ['javascript', 'node'],
-		'C plus plus': ['c++'],
 		'Cplusplus': ['c++'],
 		'xml': ['xml'],
 		'angular': ['javascript'],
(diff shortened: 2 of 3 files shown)


joaomoreno:
There are two problems:

1. You must escape the backslashes in `new RegExp('\b' + k + '\b', 'gi')`, thus becoming `new RegExp('\\b' + k + '\\b', 'gi')`.
2. You changed the description in the test to be `This C++ extension [...]`, but there is no `C++` key in the `Keywords` map, so the packaging won't detect it. Should it just be added?

waderyan:
Great. Fixed "\\b" issue. 

I'm still having a problem with matching c++. I put it back in the Keywords map and introduced a regex escape method (for the "+" character). Here is what I am seeing. 

image

And without word boundary

image


What am I missing? That should work. I introduced the word boundaries to solve this issue.

image
 

image

waderyan:
Cool - I got the answer on [SO](http://stackoverflow.com/questions/37866277/javascript-regex-matching-word-in-sentence-with-character). 

`+` is not considered a word character.

waderyan:
Fixed. Tests passing. Please review :)

joaomoreno:
Good job!

joaomoreno:
Published in `vsce@1.7.0`

PR #109, opened 2016-07-19
Display project information in Marketplace

### LINKED ISSUE #108: Support for displaying project information on Marketplace
We've added support for displaying GitHub project details and Badges on the item details page on the marketplace. But for the page to display the details we need changes to the way the tool packages the vsixmanifest. The new experience is as follows

github-proj

Following are the changes required:

- If the repository type is "Git", change the link type to Microsoft.VisualStudio.Services.Links.GitHub as opposed to Microsoft.VisualStudio.Services.Links.Repository

- We now have support for an additional badges section:

which needs to be translated to badges section as highlighted below

manifest-changes

joaomoreno:
Fixes #108

### DIFF
--- a/resources/extension.vsixmanifest
+++ b/resources/extension.vsixmanifest
@@ -11,7 +11,11 @@
       <% if (links.repository) { %>
         <Property Id="Microsoft.VisualStudio.Services.Links.Source" Value="<%- links.repository %>" />
         <Property Id="Microsoft.VisualStudio.Services.Links.Getstarted" Value="<%- links.repository %>" />
+        <% if (links.github) { %>
+        <Property Id="Microsoft.VisualStudio.Services.Links.GitHub" Value="<%- links.github %>" />
+        <% } else { %>
         <Property Id="Microsoft.VisualStudio.Services.Links.Repository" Value="<%- links.repository %>" />
+        <% } %>
       <% } %>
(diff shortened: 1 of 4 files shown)


joaomoreno:
@hamenon Done, feel free to review.

Ended up going with the following syntax for our `package.json` file:

```typescript
type badge = {
  url: string;
  href: string;
  description: string;
};
type badges = badge[];
```

hamenon:
@joaomoreno any particular reason to go with url and href instead of just link and imgUri in the json? It would help keep things consistent between the vsixmanifest and the json. We do that on the VSTS side of things with that manifest, as attached in the issue description.

arvindkrms:
@joaomoreno Can you please confirm that the following snippet in the json

``"repository": {
    "type": "git",
    "url": "https://github.com/Microsoft/vcode-spell-check.git"
}``
would be translated to 
<Property Id="Microsoft.VisualStudio.Services.Links.GitHub" Value="https://github.com/Microsoft/vcode-spell-check.git"> in the manifest file. This will make sure that we keep things consistent across VSTS and VSCode

joaomoreno:
@hamenon My preference is towards this format. We wouldn't use `uri` since it's nowhere used in the current manifest, so we'd switch it to `url` anyway. Then I thought switching it a bit would easily lead to switching it for good. The vsixmanifest also already has little to do with Code's extension manifest.

@arvindkrms The test `should detect github repositories` created in this PR proves exactly that.

hamenon:
@joaomoreno Sure, I get what you're saying. I think we can go ahead with the change.

Issue #126, opened 2016-09-15
Create a new asset type for the Changelog file


joaomoreno:
Similar to the Details asset type

joaomoreno:
@XVincentX you can also contribute to getting this done.

Check out how VSCE maps README.md into the Details asset type. We gotta make the same for the CHANGELOG.md.

XVincentX:
@joaomoreno Cool - I'll have a look later on. Stay tuned!

XVincentX:
@joaomoreno 
From a quick view I'd say that [this](https://github.com/Microsoft/vscode-vsce/blob/9871b82a4237db00f0ba8abf3d094ee259ec283e/src/package.ts#L249) is our guy to modify; probably on the same way I did in the previous PR - that is - generalise it as a `Markdown` processor and probably provide a property where you could set the regexp to find the files.

Does that make sense for you?

joaomoreno:
Spot on. 👍

joaomoreno:
Thanks @XVincentX !

PR #129, opened 2016-09-16
Make vsce recognise changelog.md as an asset

XVincentX:
The following PR will address #126 that is - if a changelog file is present in your directory, it will make sure to treat it as an asset (which can be used around the ecosystem).

Notes are in the code!

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -246,12 +246,12 @@ export class TagsProcessor extends BaseProcessor {
 	}
 }
 
-export class ReadmeProcessor extends BaseProcessor {
+export class MarkdownProcessor extends BaseProcessor {
 
 	private baseContentUrl: string;
 	private baseImagesUrl: string;
 
-	constructor(manifest: Manifest, options: IPackageOptions= {}) {
+	constructor(manifest: Manifest, options: IPackageOptions= {}, private fileRegExp : RegExp) {
 		super(manifest);
 
 		const guess = this.guessBaseUrls();
@@ -263,7 +263,7 @@ export class ReadmeProcessor extends BaseProcessor {
 	onFile(file: IFile): Promise<IFile> {
 		const path = util.normalize(file.path);
 
-		if (!/^extension\/readme.md$/i.test(path)) {
+		if (!this.fileRegExp.test(path)) {
 			return Promise.resolve(file);
 		}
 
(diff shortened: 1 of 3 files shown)


src/package.ts
```diff
@@ -246,12 +246,12 @@ export class TagsProcessor extends BaseProcessor {
...
 }
 
-export class ReadmeProcessor extends BaseProcessor {
+export class MarkdownProcessor extends BaseProcessor {
 
 	private baseContentUrl: string;
 	private baseImagesUrl: string;
 
-	constructor(manifest: Manifest, options: IPackageOptions= {}) {
+	constructor(manifest: Manifest, options: IPackageOptions= {}, private fileRegExp : RegExp) {
```
XVincentX: Are you ok with the parameter order? Usually I would go with default parameters as last one, but I did not want to break the existing behaviour. You'll tell me!

src/test/package.test.ts
```diff
@@ -195,27 +195,29 @@ describe('toVsixManifest', () => {
...
-
-		return _toVsixManifest(manifest, files)
-			.then(xml => parseXmlManifest(xml))
-			.then(result => {
-				assert.equal(result.PackageManifest.Assets[0].Asset.length, 2);
-				assert.equal(result.PackageManifest.Assets[0].Asset[1].$.Type, 'Microsoft.VisualStudio.Services.Content.Details');
-				assert.equal(result.PackageManifest.Assets[0].Asset[1].$.Path, 'extension/readme.md');
-			});
-	});
+	['readme', 'changelog'].forEach((fileType) => {
```
XVincentX: I just generalised the tests instead of copy paste others.

src/test/package.test.ts
```diff
@@ -1096,7 +1098,7 @@ describe('ReadmeProcessor', () => {
 		};
 
 		const root = fixture('readme');
-		const processor = new ReadmeProcessor(manifest);
+		const processor = new MarkdownProcessor(manifest, {}, /^extension\/readme.md$/i);
```
XVincentX: Here you can see that the parameter order is a bit...unfortunate. Suggestions?

XVincentX:
I might probably need to rebase since you released 1.12 @joaomoreno

joaomoreno:
It's alright, leave it to me.

XVincentX:
Thanks!

XVincentX:
@joaomoreno Is there anything blocking here? I'd love to move forward!

XVincentX:
@joaomoreno That's great. What would be the next step now?

joaomoreno:
Now let's go back to Microsoft/vscode#11940

PR #139, opened 2016-11-10
check for enableProposedApi and don't allow to publish

jrieken:
We want to be defensive when it comes to proposed API and one part of this is to not allow extensions to be published unless `enableProposedApi` is off.


joaomoreno:
Published in 1.17.0

PR #159, opened 2017-03-18
Fix manifest overriding

### LINKED ISSUE #158: Package.json gets overwritten with translations when the version is set to be updated automatically
It seems that the issue is related to the fact that the function `readManifest` is [always parsing the translations](https://github.com/Microsoft/vscode-vsce/blob/master/src/package.ts#L471).

Whenever the version is bumped automatically by the tool a write is going to eventually happen as you can see [here](https://github.com/Microsoft/vscode-vsce/blob/master/src/publish.ts#L112). If the version is the specified version is the same that the manifest already had then nothing happens. But if it happens to be different (which will be the case when one decides that the tool is going to do its magic) then the `writeManifest` function is called by passing the parsed manisfest (with the translations applied).

In the case of our extension this is messing with our publishing process as we check out the `package.json`'s integrity and pass some tests prior to publication.

I've prepared a PR with maybe a naive implementation. Let me know if that works. The idea is to provide an option for the `readManifest` function to only read the manifest without applying the translation parsing. Then in the versionBump we're using this to do the overwriting without issues.

robertohuertasm:
Fixes #158 

I've prepared a PR with maybe a naive implementation. Let me know if that works. The idea is to provide an option for the readManifest function to only read the manifest without applying the translation parsing. Then in the versionBump we're using this to do the overwriting without issues.

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -442,7 +442,7 @@ export function validateManifest(manifest: Manifest): Manifest {
 	return manifest;
 }
 
-export function readManifest(cwd = process.cwd()): Promise<Manifest> {
+export function readManifest(cwd = process.cwd(), nls = true): Promise<Manifest> {
 	const manifestPath = path.join(cwd, 'package.json');
 	const manifestNLSPath = path.join(cwd, 'package.nls.json');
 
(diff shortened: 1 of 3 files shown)


JimiC:
I can't shake the thought that this PR is invalid if the devs have made the following assumptions:

- If you have a single `package.nls.json` that translations should be merged with the `package.json`.
- If you have multiple translation files then you should use the `package.nls.*.json` pattern (even for English like `package.nsl.en.json`), and strictly not use a `package.nls.json` file.

robertohuertasm:
@JimiC, they will say. What I see clearly is that the code doesn't make their intentions clear. My assumption is that they only want to modify the `package.json` in order to upgrade the manifest version. If that's the case, then I think this PR is valid. From my point of view, it doesn't make sense any other kind of modification.

joaomoreno:
Great catch and fix, with tests and all! Many thanks! 👍

PR #193, opened 2017-07-14
Fixed javascript typo

t-sauer:
A lot of extensions are published with the keyword `javsacript`  and I think this is the reason why.


joaomoreno:
Oh man, great catch, thanks!

PR #215, opened 2017-11-16
Feature/issue 214 add show command

raix:
This pr. contains the initial proposal for a “show” extension metadata command

Usage:
```bash
$ vsce show <extensionId>

# or as json

$ vsce show <extensionId> --json
```
image
image

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,6 +1,7 @@
 import * as program from 'commander';
 import { packageCommand, ls } from './package';
 import { publish, list, unpublish } from './publish';
+import { show } from './show';
 import { listPublishers, createPublisher, deletePublisher, loginPublisher, logoutPublisher } from './store';
 import { getLatestVersion } from './npm';
 import { CancellationToken, isCancelledError } from './util';
@@ -111,6 +112,12 @@ module.exports = function (argv: string[]): void {
 		.command('logout <publisher>')
 		.description('Remove a publisher from the known publishers list')
 		.action(name => main(logoutPublisher(name)));
+
+	program
+		.command('show <extensionIdentifier>')
+		.option('--json', 'Output data in json format', false)
+		.description('Show extension metadata')
+		.action((extensionIdentifier, { json }) => main(show(extensionIdentifier, json)));
 
 	program
 		.command('*')
--- a/src/main.ts
+++ b/src/main.ts
@@ -114,10 +114,10 @@ module.exports = function (argv: string[]): void {
 		.action(name => main(logoutPublisher(name)));
 
 	program
-		.command('show <extensionIdentifier>')
+		.command('show <extensionid>')
 		.option('--json', 'Output data in json format', false)
 		.description('Show extension metadata')
-		.action((extensionIdentifier, { json }) => main(show(extensionIdentifier, json)));
+		.action((extensionid, { json }) => main(show(extensionid, json)));
 
 	program
 		.command('*')
--- a/src/show.ts
+++ b/src/show.ts
@@ -72,7 +72,6 @@ const indentRow = (row: string) => `  ${row}`;
 
 function showOverview({
 	displayName,
-	extensionId,
 	extensionName,
 	shortDescription,
 	versions,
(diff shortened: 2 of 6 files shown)


src/main.ts
```diff
@@ -113,6 +114,12 @@ module.exports = function (argv: string[]): void {
 		.action(name => main(logoutPublisher(name)));
 
 	program
+		.command('show <extensionIdentifier>')
```
joaomoreno: Let's use `extensionid`, just like `unpublish`.

src/show.ts
```diff
@@ -0,0 +1,134 @@
...
+export type ViewTable = ViewTableRow[];
+
+export function show(extensionUniqIdentifier: string, json: boolean = false): Promise<any> {
+	const [extensionPublisher, extensionName] = extensionUniqIdentifier.split(/\.(.+)?/, 2);
+	const flags =
+		ExtensionQueryFlags.IncludeCategoryAndTags |
+		ExtensionQueryFlags.IncludeMetadata |
+		ExtensionQueryFlags.IncludeStatistics |
+		ExtensionQueryFlags.IncludeVersions;
+	return getExtension(extensionPublisher, extensionName, undefined, flags)
```
joaomoreno: We gotta do some error handling here.

image

Something like `Extension 'joao' not found` would be better.

src/show.ts
```diff
@@ -0,0 +1,134 @@
...
+	install: number;
+	averagerating: number;
+	ratingcount:number;
+}
+
+export type ViewTableRow = string[];
+export type ViewTable = ViewTableRow[];
+
+export function show(extensionUniqIdentifier: string, json: boolean = false): Promise<any> {
+	const [extensionPublisher, extensionName] = extensionUniqIdentifier.split(/\.(.+)?/, 2);
```
joaomoreno: Kind of a strange regex here. You just need `.split('.')`.

raix: I assumed that name could contain dots, found https://github.com/Microsoft/vscode-vsce/blob/master/src/publish.ts#L185 :)

src/show.ts
raix: yeah - also taking into account that it doesn't work for you, would be nice to have more debugging.

raix: +options

src/show.ts
raix: After second thought, `extensionName` could have dots - but it would not be valid. eg.: `vsce show mortenhenriksen.perl-debug.foo` would return the package `perl-debug`. I've made the `extensionId` "parsing" more readable so it throws on invalid extension names.

src/show.ts
raix: @joaomoreno I'm wondering why it errors `Invalid Resource` could you do `npm list vso-node-api` to see what version of the api you are on?
*(I'm at `vso-node-api@6.1.2-preview`)*

src/show.ts
joaomoreno: Publishers and extension names can't have dots. 👍

joaomoreno:
@raix Got the same `vso-node-api@6.1.2-preview` dependency. Hmmm.

raix:
@joaomoreno I've added error handler for the "Invalid Resource" error with the current constraints, would be nice if the public api was available in `vso-node-api`.

joaomoreno:
Yup, that was it! I had an old access token. Interesting, I wonder if we can get this to work without depending on login.

raix:
We could just make a [post call](https://github.com/Microsoft/vscode/blob/master/src/vs/platform/extensionManagement/node/extensionGalleryService.ts#L423) directly to the [public api v3.preview](https://github.com/Microsoft/vscode/blob/b00945fc8c79f6db74b280ef53eba060ed9a1388/product.json#L18)

raix:
@joaomoreno I've added `PublicGalleryAPI` this doesn't require a pat, let me know what you think

*(one could ague that it belongs in vso-node-api)*

joaomoreno:
Great stuff. Thanks! 🍻

joaomoreno:
@modigrv Is there any way we can use the `vso-node-api` in an unauthenticated way? I've tried passing a no-op `AuthHandler` to the [`WebApi`](https://github.com/Microsoft/vsts-node-api/blob/master/api/WebApi.ts#L91) but the extension query request eventually fails while getting [versioning data](https://github.com/Microsoft/vsts-node-api/blob/master/api/GalleryApi.ts#L945).

raix:
*( Currently using the public [/_apis/public](https://github.com/Microsoft/vscode-vsce/blob/master/src/publicgalleryapi.ts#L25) the [/gallery/extensionquery](https://github.com/Microsoft/vscode-vsce/blob/master/src/publicgalleryapi.ts#L37) - similar to whats done in vs code)*

PR #220, opened 2017-12-05
fixes #219

### LINKED ISSUE #219: vsce show fails when manifest has no tags property
testing https://github.com/Microsoft/vscode/issues/39586

`vsce show hoovercj.vscode-power-mode`
Result: Error: Cannot read property 'join' of undefined

chrisdias:
there are a couple of ways to protect this scenario, not sure if this is the best way but it works.


joaomoreno:
Thanks! 🍻

PR #223, opened 2017-12-05
Hide file extension tags in "show" command

### LINKED ISSUE #222: vsce show prints __tags
**repo**
1. Run `vsce show CristianRengifo.informix4gl`

**bug**
The following tags are printed: 

```
  __ext_4gl, __ext_per, 4GL, Informix, Informix 4GL
```

I don't know where these double underscore tags are coming from. Probably file extension support?

Here's the package.json on the marketplace:

screen shot 2017-12-05 at 11 31 39 am

raix:
Tags may contain file extensions prefixed `__ext_` we filter those out from the tags list

* added file extension tag filter
* guard show arguments with default values

fixes #222


joaomoreno:
Makes sense!

PR #225, opened 2017-12-06
fix "show" to be case insensitive

### LINKED ISSUE #224: VSCE Show - Windows output hard to read
Testing https://github.com/Microsoft/vscode/issues/39586

Two minor issues:
1. The Windows shell struggles with the font for ratings and download
2. Perhaps we should state that extension names can be case sensitive (I know they are meant to be all lower case but some are not)

image

raix:
The `extensionId` should not be case sensitive. *Matching a search on the Marketplace.*

fixes #224


joaomoreno:
Thanks again! 🍻

PR #226, opened 2017-12-06
fix show: add icon fallback on win32

### LINKED ISSUE #221: vsce show on powershell/cmd prints unknown characters
**Repo steps**
1. In windows powershell or cmd
2. Run `vsce show eamodio.gitlens`

**bug**
The stars and download symbol are printed as unknown characters

capture

raix:
Both cmd and powershell fails to render the super nice unicode stars. It breaks my hearts... to make windows users use a fallback.

The issue circles unicode glyphs not in raster etc. fonts and the use of charpages.

fixes #221


joaomoreno:
Thanks! 🍻

PR #229, opened 2017-12-14
Add execute permission for files in vsix

dlebu:
Mac binaries packaged to a vsix on Windows were not executable when installing vsix on Mac. (mode 664). This change makes all files executable by the owner as well (mode 764). This might be an issue in yazl, since [they say](https://github.com/thejoshwolfe/yazl#addfilerealpath-metadatapath-options) they respect the original `stats.mode` of a file. In practice, the mode is not respected when you compare before/after creating vsix modes. @srivatsn

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -692,8 +692,8 @@ function writeVsix(files: IFile[], packagePath: string): Promise<string> {
 			const zip = new yazl.ZipFile();
 			let options = {
 				mtime: new Date(),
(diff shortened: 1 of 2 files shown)


joaomoreno:
@dlebu This just adds the executable bit for every file within the extension. We can't merge that, it's a security problem.

At best we can come up with a whitelisting feature in which you can describe which files should have the executable bit.

PR #248, opened 2018-03-24
if package path is a directory, use default file name to write vsix

LaChRiZ:
I improved the package path option.

If the given package output path is just a directory, it will be concatenated with the default file name (`{manifest.name}-${manifest.version}.vsix`) to write the VSIX file.
If it's a concrete file path, that will be used.

Otherwise, the default behaviour will be used. (current working directory with package name and version)

Hope you like it. Thanks for merging! 😊

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -822,6 +822,17 @@ function defaultPackagePath(cwd: string, manifest: Manifest): string {
 	return path.join(cwd, `${manifest.name}-${manifest.version}.vsix`);
 }
 
+function getPackagePath(packagePath: string, cwd: string, manifest: Manifest): string {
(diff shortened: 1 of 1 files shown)


garyng:
Uhmm... Will this be merged?

LaChRiZ:
I rebased the feature branch. So it could me merged now.
I would appreciate it. It's a handy feature.

joaomoreno:
I find this very confusing. I think there should really be a check for whether a directory exists. If so, that directory is taken as a base. Otherwise, we should take that name literally as the package name. I'll rebase and merge it.

joaomoreno:
Thanks! :fireworks:

PR #260, opened 2018-05-03
Support Yarn use in package command

kevcunnane:
This fixes an issue where `npm list` thinks dependencies are invalid,
but a project uses Yarn and this can correctly handle the dependencies.
This is blocking SQL Ops Studio extensions that rely on
https://github.com/anthonydresser/dataprotocol-client, as this is marked
as invalid and essentially requires use of Yarn.

Please let me know if there's anything I should be doing in terms of testing etc. to help validate this. The basic test of using `vsce package --yarn` worked as expected and no longer caused packaging to fail.


joaomoreno:
Looks great, thanks! v1.40.0 is on the way of being built: https://travis-ci.org/Microsoft/vscode-vsce/builds/374278813. Let me know how it goes for you.

PR #262, opened 2018-05-24
Pull minimum translations out to manifest

ramya-rao-a:
Pull the [translations](https://github.com/Microsoft/vscode/blob/57cca7d272de3cef41bf8239f1547a39f82934c9/src/vs/platform/node/minimalTranslations.ts) needed to prompt user to install language pack to package.json so that they can be used by vscode without downloading the whole extension

Related to Microsoft/vscode#45663

cc @sandy081 @dbaeumer

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -81,6 +81,20 @@ export class BaseProcessor implements IProcessor {
 	onEnd() { return Promise.resolve(null); }
 }
 
+interface ILocalization {
+	languageId: string;
+	translations: ITranslation[];
+	mainTranslationPath: string;
+	minimalTranslations: { [key: string]: string };
+}
+
+interface ITranslation {
+	id: string;
+	path: string;
+}
+
+let localizations: ILocalization[];
+
 function getUrl(url: string | { url?: string; }): string {
 	if (!url) {
 		return null;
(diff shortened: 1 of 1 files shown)


src/package.ts
```diff
@@ -531,6 +545,37 @@ class IconProcessor extends BaseProcessor {
...
+			const contentsJson = JSON.parse(contents);
+			currentLoc.minimalTranslations = contentsJson["contents"] ? contentsJson["contents"]["vs/platform/node/minimalTranslations"] : null;
+		}
+
+		return file;
+	}
+
+	onEnd(): Promise<void> {
+		if (localizations.some(x => !x.minimalTranslations)) {
+			return Promise.reject(new Error(`At the bare minimum, translations for the "vs/platform/node/minimalTranslations" should be present in the extension.`));
```
joaomoreno: Should this inform which language lacks the `minimalTranslations`?

sandy081:
I think polluting `package.json` with generated stuff is not a good idea. My initial suggestion was to put the minimal translation strings in `.vsixmanifest` file as a property similar to `extensionDependencies` property. Sorry if I was not clear enough about that.

After discussing with @joaomoreno we came up with a better approach. Expose the translation file that contains minimal translation strings as an asset which can be extracted from VS Code. VS Code can then get the minimal translation strings from that asset file. This way we do not generate any in vsce.

joaomoreno:
Closing as per @sandy081's comment.

1.41.0 is about to be release with the following change: https://github.com/Microsoft/vscode-vsce/commit/f4d50e207315db21a0e6ac13fb8117406fe89bee

ramya-rao-a:
Thanks

Issue #180, opened 2017-06-18
Support version bump for package-lock.json when publishing


JimiC:
For those using `npm 5`, `package-lock.json` is added in the project, which also contains a `version` property.

Whenever publishing an extension, the version should be bumped in `package.json` and `package-lock.json` to keep them in sync.

joaomoreno:
Wanna try submitting a PR?

JimiC:
Yeah, I tried looking at the code but didn't felt confident tampering with it. I believe @robertohuertasm has more experience with it and maybe he can assist.

felixfbecker:
Why is it vsce's responsibility to bump versions? Why not use `npm version`?

JimiC:
@felixfbecker Because `vsce` is bumping the `package.json` version on publish (https://github.com/Microsoft/vscode-vsce/blob/83060c333a56aac26019df204af4ae948c041326/src/publish.ts#L87)

felixfbecker:
I understand that vsce provides this feature, but I am questioning whether it should be as it is already provided by npm itself

JimiC:
@felixfbecker Could you please elaborate? I'm not so familiar with `npm 5` and its features, as it's fairly new and still evolving.

JimiC:
@felixfbecker No need to elaborate. I figured out what you want to say. Still though, although you can bump the version of `package.json` and `package-lock.json` with `npm version`, `vsce` provides the ability to bump the version from the tool. And while the tool does bump the version of `package.json` it should take into consideration the existence of `package-lock.json` file as well.

felixfbecker:
Imo the functionality should just be deprecated. `npm version` is guaranteed to always work with all lockfiles and shrinkwrap, it can also commit and tag the new version. So vsce doesn't add value here, but inevitably will have bugs like this issue that need work over time to fix

JimiC:
@felixfbecker That's an entirely new proposal. Maybe you should open an issue and promote it.

/cc @joaomoreno

joaomoreno:
I have to agree with @felixfbecker. We should reuse as much as possible. We might just deprecate this behaviour from vsce altogether.

Closing this issue as it becomes obsolete.

JimiC:
Good to know. Precious development time saved.

PR #263, opened 2018-06-02
Add ability to publish using  'npm version'

### LINKED ISSUE #206: Preserve existing indentation in package.json
Using `vsce publish <version>` is clobbering the existing indentation in `package.json` and forcing it to always be 4 spaces. This differs from the default 2-space indentation generated by `npm init`, and can make git diffs rather messy.

I know `npm` has code to detect the indentation of `package.json` and preserve it when installing/updating/removing dependencies or publishing a new version. Ideally `vsce` would do the same.

I’ve been able to quickly get the same functionality working for `vsce` locally (using the same module as `npm` for consistency). I’d be happy to create a PR for it, but I also saw in #180 that `vsce` might be moving away from custom `package.json` updates completely. So before I clean up the code and write tests for it, is this a PR that’s worth doing, or should I just wait until `vsce` uses `npm publish` under the covers?

JimiC:
Resolves #206 
References #180 

As per popular request `vsce publish <version>` will now use `npm version <version>`, in order to bump the package version. In the process, `package-lock.json` will also be updated (when using `npm5+`) and tagged `git` commit will be done (supporting `gpg` signing when it's set in `git` config).

**Note:** As before only `major`, `minor` and `patch` are supported as a valid ` <version>`.

I refrained from implementing a rollback of the process, as if anything goes wrong after the version bump (i.e. network issues), the user can simply execute `vsce publish` (without ` <version>`) and complete the publishing.

This implementation also does not interfere with using plain `vsce publish` via a CI service. Tthe user, in that case, should perform `npm version <version>` locally. Then push the commit and the tag. The CI publishing stage should only call `vsce publish -p <token>`.

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -23,7 +23,6 @@ interface IReadFile {
 }
 
 const readFile = denodeify<string, string, string>(fs.readFile);
-const writeFile = denodeify<string, string, string, void>(fs.writeFile as any);
 const unlink = denodeify<string, void>(fs.unlink as any);
 const exec = denodeify<string, { cwd?: string; env?: any; }, { stdout: string; stderr: string; }>(cp.exec as any, (err, stdout, stderr) => [err, { stdout, stderr }]);
 const glob = denodeify<string, _glob.Options, string[]>(_glob);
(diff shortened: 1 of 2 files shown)


joaomoreno:
Works great for me!

Issue #267, opened 2018-06-20
How to publish extension when we're using yarn?


ritwickdey:
I can build the package using `vsce package --yarn`, but when I can't do `vsce publish --yarn`. 

I'm switching to yarn for https://github.com/Microsoft/vscode-vsce/issues/246

joaomoreno:
Wanna contribute a PR?

ritwickdey:
Yep! sure

JimiC:
This issue should be closed.

joaomoreno:
Thanks!

PR #268, opened 2018-06-21
useYarn parameter while publishing extension

ritwickdey:
`--yarn` option is added with `vsce publish` command.

Issue #267


joaomoreno:
Coming out in v1.43.0.

PR #271, opened 2018-07-03
Fix pattern matching in .vscodeignore

### LINKED ISSUE #265: .vscodeignore fails to exclude folders when they do not have a slash at the end
When you have a folder name in your .vscodeignore without a / or /** at the end vsce includes it, but it should exclude it.

JimiC:
Fixes #265 

This PR attempts to keep backward compatibility and be compliant with `Git` ignoring pattern format ([`.gitignore`](https://git-scm.com/docs/gitignore#_pattern_format))

Test fixtures and case have been included.

Note: `fs.lstat` has been used instead of `fs.exists` ([line 708](https://github.com/Microsoft/vscode-vsce/pull/271/files#diff-3e810d99c2069fa4c4d8bd38e3286e21R708)), in order to be compliant with deprecation of `fs.exists` in `node@10.x`. `fs.existsSync` was not used deliberately, in order to avoid I/O blocking.


JimiC:
@joaomoreno Is there something blocking the merging?

gulpfile.js
```diff
@@ -34,8 +34,8 @@ function compile() {
 	const api = ts.dts
 		.pipe(filter('**/api.d.ts'));
 
-	const resources = gulp.src('src/**')
-		.pipe(filter(['**', '!**/*.ts']));
+	const resources = gulp.src('src/**', { dot: true })
+		.pipe(filter(['**', '!**/*.ts'], { dot: true }));
```
joaomoreno: Does this actually change anything?

gulpfile.js
```diff
@@ -54,7 +54,7 @@ gulp.task('test', ['compile'], test);
 gulp.task('just-test', ['just-compile'], test);
 
 function watch(task) {
-	return cb => gulp.watch(['src/**', 'typings/**'], [task]);
+	return () => gulp.watch(['src/**', 'typings/**'], [task]);
```
joaomoreno: This is wrong since gulp checks for the args `length` of the function. If there is no parameter in the function, gulp expects a stream. If it doesn't return a stream, it marks the task as immediately completed. I suggest to leave the `cb` there.

src/test/package.test.ts
```diff
@@ -135,6 +135,39 @@ describe('collect', function () {
...
+		}
+
+		if (!fs.existsSync(path.join(cwd, 'out', 'hello'))) {
+			fs.writeFileSync(path.join(cwd, 'out', 'hello'), 'world');
+		}
+
+		// pattern of file name
+		if (!fs.existsSync(path.join(cwd, 'logger.log'))) {
+			fs.writeFileSync(path.join(cwd, 'logger.log'), 'log');
+		}
```
joaomoreno: We should create a fixture instead of generating it dynamically here, it would be easier to grasp and maintain.

src/package.ts
```diff
@@ -701,7 +702,15 @@ function collectFiles(cwd: string, useYarn = false): Promise<string[]> {
...
-			.then(rawIgnore => rawIgnore.split(/[\n\r]/).map(s => s.trim()).filter(s => !!s))
+			.then(rawIgnore => rawIgnore.split(/[\n\r]/).filter(s => !!s).map(s => s.trim().replace(/\/$/, '')))
+			.then(ignore => {
+				const promises = ignore.map(i => {
+					return lstat(i)
+						.catch<fs.Stats>(() => Promise.resolve())
+						.then(stat => stat && stat.isDirectory() ? `${i}/**` : i);
+				});
+				return Promise.all(promises).then<string[]>(i => Promise.resolve(i));
+			})
```
joaomoreno: Is there really not a better way to do this without disk access?

joaomoreno:
> @joaomoreno Is there something blocking the merging?

Only my vacation :sunglasses:

gulpfile.js
JimiC: Have abandoned `gulp` years ago, so I have to admit that I don't remember how it works. Since it's crucial I'll revert this.

src/test/package.test.ts
JimiC: My original thought but I tried to keep the coding pattern. Will switch to fixture files.

src/package.ts
JimiC: I can't think of another way to determine if an entry is a directory or a file. Suggestions are welcome.

joaomoreno:
Thanks! 🎆 

I removed the `lstat` usage by simply detecting whether a pattern is possibly a folder and then appending `${folder}/**` to the ignore list. That should work for most cases. We let glob do all the FS work.

JimiC:
🙄 Interesting approach. Since it works, fine by me.

jedwards1211:
I guess no one was aware of the `ignore` npm package?

Issue #273, opened 2018-07-12
Support extension packs


sandy081:
New approach of extension packs will define extensions using `extensionPack` property. VSCE should populate this as a manifest property in VSIX just like `extensionDependencies`

PR #274, opened 2018-07-12
Populate extension pack property

sandy081:
Implements #273


joaomoreno:
Coming out in 1.44.0.

PR #280, opened 2018-08-19
Changed --out flag description in package command

amadeann:
I was quite confused by the previous description ('Location of the package'). I think it's worth being more explicit about what is being created.


joaomoreno:
Makes a lot of sense. Thanks! :beers:

PR #281, opened 2018-08-20
Show URL of published extension when it is successful

sri:
screen shot 2018-08-19 at 10 36 24 pm

Provide some guidance to the user as to the URL of the plugin.


joaomoreno:
Thanks! :beers:

PR #282, opened 2018-08-30
Support for 'vscode:packagedDependencies'

jrieken:


joaomoreno:
:thinking: Tests seem to fail. @jrieken

jrieken:
Yeah, looks like no yarn on the build machine?

jrieken:
Removed the package.json-trick and make `packagedDependencies: string[]`-api for the `listFiles` function

joaomoreno:
Published `vsce@1.47.0`

PR #286, opened 2018-09-14
Add property for supported languages

ramya-rao-a:
Related to https://github.com/Microsoft/vscode/pull/58563

This PR adds a new property to the vsixmanifest that contains the localized names of the languages supported by the extension.

PR #287, opened 2018-09-17
Use localizedLanguageName instead of languageNameLocalized

ramya-rao-a:
This is a follow up to #286

We have been using the wrong property name there.


joaomoreno:
Published 1.49.2

PR #294, opened 2018-10-02
Ignore file via option

### LINKED ISSUE #255: Option to use a particular .vscodeignore file
Is there an option to use a particular .vscodeignore file (which may not be present directly in the current working directory or one with a different name) ?

alefragnani:
Hi,

This PR adds `--ignoreFile [path]` option to `ls`, `package` and `publish` commands (Closes #255 ).

The code changes where minimal. I simply added the new parameter, following the idea of the already existing ones. Let me know if any updated needs to be made

Thanks


joaomoreno:
Thanks and sorry for the delay! :fireworks:

Issue #230, opened 2017-12-27
.vsce file permissions world-readable by default


cdyson37:
I've just used vsce for the first time, so apologies in advance if I've got this wrong. The `.vsce` file appears to contain a secret access token, but it appears to have been created with my regualar umask (I'm on Linux) and is thus world-readable. I think if in doubt it should be readable only by the creating user.
```
$ ls -l .vsce
-rw-r--r-- 1 charlie charlie 97 Dec 27 17:57 .vsce
```

PR #295, opened 2018-10-09
Set permissions on .vsce to 0600; fix for #230

ndrake:
This won't fix permissions if .vsce already exists, so maybe this isn't a complete fix?


joaomoreno:
Thanks for the PR!

PR #312, opened 2018-12-14
Respect Yarn choice when publishing

jsayol:
When the `--yarn` option is specified for publishing, respect its usage over npm.


joaomoreno:
This has been fixed in `master` since. Sorry!

PR #330, opened 2019-02-06
Index color and icon themes

sandy081:
Index color and icon themes so that they can be previewed in VS Code without downloading the vsix


sandy081:
@joaomoreno As discussed created an asset file in the root for all colour and icon themes respectively.

Issue #192, opened 2017-07-13
Relative sources in <img> tags in README.md are not rewritten to absolute URLs


fabiospampinato:
I wrote a readme file in markdown, and I had no problems putting a local gif in it, but vsce complained about a local png, maybe because I inserted it like this (just for centering it):

```md
<p align="center">
	<img src="resources/logo.png" alt="Logo">
</p>
```

joaomoreno:
Can you show me what "complained" means? Did you get an error?

fabiospampinato:
Exactly, more specifically this error:
```
Error: Images in README.md must come from an HTTPS source: resources/logo.png
```

joaomoreno:
Is your extension in a GitHub repository?

fabiospampinato:
Sure, [here](https://github.com/fabiospampinato/vscode-projects-plus) it is.

joaomoreno:
I've cloned that repo and ran `npm i` and then `vsce package`. It worked:

```
vscode-projects-plus master* → vsce package
Executing prepublish script 'npm run vscode:prepublish'...

> vscode-projects-plus@1.0.1 vscode:prepublish /Users/joao/Desktop/vscode-projects-plus
> tsc -p ./

Created: /Users/joao/Desktop/vscode-projects-plus/vscode-projects-plus-1.0.1.vsix
```

Something must be off in your machine.

Do you have the latest vsce installed?

fabiospampinato:
You forgot to edit the readme, it worked because now it's written in other way, otherwise I wouldn't be able to publish it because of the error I'm reporting.

Replace this:
```md
<img src="https://raw.githubusercontent.com/fabiospampinato/vscode-projects-plus/master/resources/logo-128x128.png" alt="Logo">
```

With this:
```md
<img src="resources/logo.png" alt="Logo">
```

If you have `sed` available just run:
```
sed -i -e 's/https:\/\/raw.githubusercontent.com\/fabiospampinato\/vscode-projects-plus\/master\/resources\/logo-128x128.png/resources\/logo.png/g' README.md
```

Yes, I'm using the latest version of vsce.

fabiospampinato:
I believe I found the problem, [here](https://github.com/Microsoft/vscode-vsce/blob/6d3a1e43d712b7f15658039517a4b297b144f530/src/package.ts#L325) it checks if the url is relative, but just for images written in the markdown syntax.

[...]

PR #335, opened 2019-03-03
img html tags in readme

dlech:
This is a quick hack to address issue #192. It adds a regex to match and modify relative URLs in `` tags, taking care to watch out for `src="data:image/..."`. Test cases have been added.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+*		text=auto
+*.gif	binary
+*.jpg	binary
+*.png	binary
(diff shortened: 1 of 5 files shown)


dlech:
Ping. I have to maintain a fork of vsce because of this. Also #208 does the same thing, so I am not the only person who finds this useful enough to make PR.

joaomoreno:
Thanks for the ping, I've decided to merge the older PR: https://github.com/microsoft/vscode-vsce/pull/208

Released in 1.69.0.

dlech:
Great. Thank you.

Issue #333, opened 2019-02-11
Recommend bundling extension


jrieken:
When publishing vsce should count the files that end up in the bundle and recommend bundling. We see extensions that load very slow (30-40 seconds) and when bundled they usually load 10 times faster. Also installing benefits a lot from this and therefore its time to spread the news!

joaomoreno:
Yeah good point. But we should probably also update the docs: https://code.visualstudio.com/api/working-with-extensions/publishing-extension

jrieken:
We are tracking that here: https://github.com/Microsoft/vscode-docs/issues/2254#issuecomment-463718248

Issue #334, opened 2019-03-01
Check for webpack


chrisdias:
VSCE should check to see if an extension is webpack'd and if not, print out a link to the (forthcoming) documents on how to webpack an extension.  this will help the ecosystem keep/make VS Code fast.

alefragnani:
That’s a great idea 👍 

I would suggest to add webpack option/configuration on [Yeoman generator](
https://github.com/Microsoft/vscode-generator-code) as well.

PR #336, opened 2019-03-04
Suggest bundling when having many files

jrieken:
This for https://github.com/Microsoft/vscode-vsce/issues/333 and suggests to bundle when having an extension with more than 100 files. The limit 100 is a bit arbitrary tbh, open for other suggestions. 50? 1000?

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -840,6 +840,9 @@ export async function pack(options: IPackageOptions = {}): Promise<IPackageResul
 	manifest = await prepublish(cwd, manifest);
 
 	const files = await collect(manifest, options);
+	if (files.length > 100) {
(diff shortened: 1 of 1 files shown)


jrieken:
Before *publishing* this version we must publish the website update

joaomoreno:
Are there any other signs we could check besides file count > 100?

@chrisdias also suggests we should check whether webpack is in the dependencies: https://github.com/Microsoft/vscode-vsce/issues/334

jrieken:
Having webpack doesn't mean you know how to use it ;-) Tho, we should check now many JS-files you have.

joaomoreno:
Right, what if it's already bundled, does the user have to read this message all the time she packages?

jrieken:
If you bundle in 100+ files then you are doing it wrong

joaomoreno:
Just ping me when this should be published.

PR #347, opened 2019-05-09
Fix compile task

JimiC:
This is a proposed PR to fix the `compile` task. Please feel free to close if not approved.


joaomoreno:
Yes! Thanks! Missed the tasks, when removing gulp.

Issue #346, opened 2019-05-07
Use an updated cpx to fix security vulnerabilities


sean-mcmanus:
vscode-vsce references https://github.com/mysticatea/cpx which has known security vulnerabilities, which have been fixed in the repo, but the owner of the cpx is unresponsive and not releasing an update with the fixes. Can someone do something like clone the repo and reference the clone instead to fix this?

kondratyev-nv:
Right now, `npm audit` is failing for extensions with `vsce >= 1.60.0` and I can not update this package without disabling the `npm audit` step in CI :(

I saw that `cpx` was introduced in https://github.com/microsoft/vscode-vsce/commit/0c517649bee2273465e7f05e2690982a6c334275. Does this package have to be declared in `dependencies`? It looks like it is used only for development purposes. So maybe it is possible to move it to `devDependencies`? I think it should at least fix the builds for the extensions.

joaomoreno:
Yeah moving it to dev deps should be fine for now.

PR #348, opened 2019-05-11
Move cpx package to devDependencies

kondratyev-nv:
Fixing vulnerability for extensions with `vsce >= 1.60.0` (#346).


joaomoreno:
Thanks!

PR #354, opened 2019-05-23
Check @types/vscode version against engines.vscode

octref:
See context at https://github.com/microsoft/vscode/issues/71048.

### DIFF
--- a/src/validation.ts
+++ b/src/validation.ts
@@ -59,12 +59,16 @@ export function validateVSCodeTypesCompatibility(engineVersion: string, typeVers
 
 	try {
 		const engineSemver = parseSemver(`vscode@${engineVersion}`);
-		const typeSemver = parseSemver(`@types/vscode@${typeVersion}`);
-
 		plainEngineVersion = engineSemver.version;
+	} catch (err) {
+		throw new Error('Failed to parse semver of engines.vscode');
+	}
+
+	try {
+		const typeSemver = parseSemver(`@types/vscode@${typeVersion}`);
 		plainTypeVersion = typeSemver.version;
 	} catch (err) {
-		throw new Error('Failed to parse semver of engines.vscode and @types/vscode');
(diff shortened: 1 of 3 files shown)


src/validation.ts
```diff
@@ -40,4 +41,59 @@ export function validateEngineCompatibility(version: string): void {
...
+	let plainEngineVersion: string, plainTypeVersion: string;
+
+	try {
+		const engineSemver = parseSemver(`vscode@${engineVersion}`);
+		const typeSemver = parseSemver(`@types/vscode@${typeVersion}`);
+
+		plainEngineVersion = engineSemver.version;
+		plainTypeVersion = typeSemver.version;
+	} catch (err) {
+		throw new Error('Failed to parse semver of engines.vscode and @types/vscode');
```
joaomoreno: Maybe we should have one try-catch block for each parse.

src/validation.ts
```diff
@@ -40,4 +41,59 @@ export function validateEngineCompatibility(version: string): void {
...
+	}
+
+	// For all `x`, use smallest version for comparison
+	plainEngineVersion = plainEngineVersion.replace(/x/g, '0');
+
+	const [typeMajor, typeMinor, typePatch] = plainTypeVersion.split('.').map(x => {
```
joaomoreno: Should we use `semver.parse` instead?

octref: I think no

```
> s.parse('^1.2.0')
null
> s.parse('>=1.2.0')
null
> s.parse('~1.2.0')
null
> s.parse('1.2.0')
SemVer {
  options: { loose: false, includePrerelease: false },
  loose: false,
  raw: '1.2.0',
  major: 1,
  minor: 2,
  patch: 0,
  prerelease: [],
  build: [],
  version: '1.2.0' }
> s.parse('1.2')
null
>
```

PR #359, opened 2019-06-04
Remove stub type definition for commander

jamesgeorge007:
`@types/commander` package is deprecated.

commander provides its own type definitions and it is not required to install `@types/commander` as from the official [npm package repository](https://www.npmjs.com/package/@types/commander) page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
   },
   "devDependencies": {
     "@types/cheerio": "^0.22.1",
-    "@types/commander": "^2.12.2",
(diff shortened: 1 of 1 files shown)


jamesgeorge007:
cc @joaomoreno

joaomoreno:
Thanks! :beers: 

You just forgot the [`yarn.lock` file changes](https://github.com/microsoft/vscode-vsce/commit/3a54891b07327a9844f0caf0c9d96965e93635dc).

PR #363, opened 2019-06-14
feat(chore): Added update-notifier

jamesgeorge007:
Included the provision to inform users of the updates in a non-intrusive way.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "semver": "^5.1.0",
     "tmp": "0.0.29",
     "typed-rest-client": "1.2.0",
+    "update-notifier": "^3.0.0",
     "url-join": "^1.1.0",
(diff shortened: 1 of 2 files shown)


jamesgeorge007:
@joaomoreno Thoughts :thinking:

joaomoreno:
VSCE already does this: https://github.com/microsoft/vscode-vsce/blob/master/src/main.ts#L49

jamesgeorge007:
@joaomoreno It seems that the current warning prompt showing up is just plain text. Why not have a similar approach as the case of  `update-notifer` wherein which the content is formatted using packages like `boxen`.

jamesgeorge007:
Just thoughts :thinking:

joaomoreno:
It sounds like bloat to me. Let's keep this simple.

Issue #364, opened 2019-06-17
Cannot specify commit message


nagytech:
I can see that `vsce publish` discretely uses `npm version` to apply versioning to `package.json` which also adds a commit  to the local repository.  The default commit message turns out to be '[major].[minor].[patch]', and there is currently no way to override this.

When using `vsce publish` as a part of a CI/CD pipeline, I need to be able to specify a commit message that  includes an ignore flag so that builds are not re-triggered by the build agent pushing changes.

nagytech:
Related to #9

joaomoreno:
@nagytech Why not running `npm version -m MESSAGE` yourself and then simply `vsce publish`?

nagytech:
Because `vsce publish` is already calling `npm version` there is an expectation that the versioning should be handled internally.

raix:
@nagytech I don't follow the title of this issue completely - but I'll share a couple of options, depending on your setup:

---

I use semantic release with the vsce plugin to automate the CI for the perl debug extension on github, takes some practice to get into - but on the other hand we get release notes/semantic versioning for free and contributors don't have to wait for me to publish.

---

It's also possible to setup a multi-branch build pipeline with a counter for the version and only publish when committing to master (additional add pre releases eg. on feature branches) or setup a release pipeline for promotion of versions manually. _(Azure Pipelines works well in this case but jenkins also supports this - and I would be surprised if travis is lagging behind)_

_(in these cases the version is tracked by the build system not package.json eg. `vsce publish ${BuildNumber}` - so no additional commits - optional tagging of the version on successful builds)_

nagytech:
The build process I'm following is as follows:

```
vsce publish minor -p $PAT   
git push origin master
```

As noted before, `vsce publish [version]` will bump the version in `package.json` _and_ add a commit to the local git repository.  The problem here is that there are no options to control the format of the commit message.  This causes cyclical builds which I'm trying to avoid.

So far, all of the options I've explored rely on external tools, processes or adding additional steps to my build.  What I really want is to be able to run the two existing build commands (with a custom commit message) and walk away without worrying about CI/CD imploding.

The changes I've already submitted help me accomplish this and are done in a way that is as unobtrusive as possible.

Hopefully that's not too much to ask.

joaomoreno:
VSCE only bumps the version because you ask it to by providing `minor`. Try this instead:

```
npm version minor -m MESSAGE
vsce publish -p $PAT
git push origin master
```

[...]

PR #365, opened 2019-06-17
Allow commit message to be specified

nagytech:
https://github.com/microsoft/vscode-vsce/issues/364


joaomoreno:
Thanks! :beers:

PR #366, opened 2019-06-19
fix(chore): Minor refactor

jamesgeorge007:
Use `process.platform` instead of `os.platform() `to maintain consistency.

### DIFF
--- a/src/viewutils.ts
+++ b/src/viewutils.ts
@@ -1,5 +1,3 @@
-const os = require('os');
-
 export type ViewTableRow = string[];
 export type ViewTable = ViewTableRow[];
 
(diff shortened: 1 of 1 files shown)


jamesgeorge007:
@joaomoreno Makes sense :thinking:

raix:
`process.platform` was added in `v0.1.16` and `os.platform()` was added in `v0.5.0` (later)
Is `os.platform()` being deprecated?

jamesgeorge007:
@raix There is no harm in having `os.platform()`. But still, the refactor ensures consistency IMO.

joaomoreno:
Sure, why not.

PR #376, opened 2019-07-19
Honors --yarn flag for prepublish step

eamodio:
Uses yarn for the prepublish step if `--yarn` is specified on the command line

/cc @joaomoreno


eamodio:
FYI, this change has a pretty big performance improvement for package/publish (at least for GitLens)

eamodio:
@joaomoreno any chance of something like this landing soon? Without it, packaging GitLens takes ~180s rather than ~68s. So I've been regularly hand patching it -- which is annoying because I wipe out `node_modules` before packaging 😄

joaomoreno:
Right now!

Issue #246, opened 2018-03-19
vsce package fails


ribrdb:
I'm trying to package my extension based on the lsp-sample. I run vsce package in the client directory and it says:
```
$ vsce package
Executing prepublish script 'npm run vscode:prepublish'...

> vscode-closure-completion-client@0.0.1 vscode:prepublish /Volumes/Seagate/Users2/ribrdb/Developer/vscode-extension-samples-master/lsp-sample/client
> ../node_modules/.bin/tsc -p ./

Error: Command failed: npm list --production --parseable --depth=99999
npm ERR! missing: block-stream@https://registry.npmjs.org/block-stream/-/block-stream-0.0.9.tgz, required by tar@2.2.1
npm ERR! missing: inherits@https://registry.npmjs.org/inherits/-/inherits-2.0.3.tgz, required by block-stream@https://registry.npmjs.org/block-stream/-/block-stream-0.0.9.tgz
```

joaomoreno:
If you run `npm list --production --parseable --depth=99999`, what happens?

marquesmps:
I am having the same problem. When I run that command I get:
```
$ npm list --production --parseable --depth=99999
npm ERR! missing: graceful-readlink@1.0.1, required by commander@2.9.0
npm ERR! missing: argparse@1.0.10, required by js-yaml@3.2.7
npm ERR! missing: esprima@2.0.0, required by js-yaml@3.2.7
npm ERR! missing: sprintf-js@1.0.3, required by argparse@1.0.10
npm ERR! missing: docker-modem@1.0.6, required by dockerode@2.5.4
npm ERR! missing: concat-stream@1.5.2, required by dockerode@2.5.4
npm ERR! missing: tar-fs@1.12.0, required by dockerode@2.5.4
npm ERR! missing: JSONStream@1.3.2, required by docker-modem@1.0.6
npm ERR! missing: debug@3.1.0, required by docker-modem@1.0.6
npm ERR! missing: readable-stream@1.0.34, required by docker-modem@1.0.6
npm ERR! missing: split-ca@1.0.1, required by docker-modem@1.0.6
npm ERR! missing: jsonparse@1.3.1, required by JSONStream@1.3.2
npm ERR! missing: through@2.3.8, required by JSONStream@1.3.2
npm ERR! missing: ms@2.0.0, required by debug@3.1.0
npm ERR! missing: core-util-is@1.0.2, required by readable-stream@1.0.34
[...]

PR #379, opened 2019-07-26
feat: proper yarn workspace support

nlunets:
Fixes https://github.com/microsoft/vscode-vsce/issues/300

This PR changes the way yarn bundling is being done. 
The main thing to consider is that `yarn list` is nowhere close to what `npm ls` does so you cannot really rely on that for bundling, especially in a workspace context.

Instead the code goes through the dependency based on the node resolution algorithm https://nodejs.org/api/modules.html#modules_all_together

At package time this means we might end up with path like "../../../node_modules" which obviously do not make sense in a packaged environment, so those path are stripped to just be node_modules ...

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -837,7 +837,14 @@ export function collect(manifest: Manifest, options: IPackageOptions = {}): Prom
 	const processors = createDefaultProcessors(manifest, options);
 
 	return collectFiles(cwd, useYarn, packagedDependencies).then(fileNames => {
-		const files = fileNames.map(f => ({ path: `extension/${f}`, localPath: path.join(cwd, f) }));
+		const files = fileNames.map(f => {
+			// In case of a yarn workspace the filepath can point to node_modules at a higher level, we can just strip that within the extension
+			let basePathF = f;
+			while(basePathF.indexOf('../') === 0) {
+				basePathF = basePathF.substr(3);
+			}
+			return ({ path: `extension/${basePathF}`, localPath: path.join(cwd, f) });
+		});
 
 		return processFiles(processors, files, options);
 	});
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -71,29 +71,39 @@ export interface YarnDependency {
 	children: YarnDependency[];
 }
 
-function asYarnDependency(prefix: string, tree: YarnTreeNode, prune: boolean): YarnDependency | null {
-	if (prune && /@[\^~]/.test(tree.name)) {
+function asYarnDependency(prefix: string, name: string, prune: boolean, parentStack:string[]): YarnDependency | null {
+	if (prune && /@[\^~]/.test(name)) {
 		return null;
 	}
 
-	let name: string;
-
-	try {
-		const parseResult = parseSemver(tree.name);
-		name = parseResult.name;
-	} catch (err) {
-		name = tree.name.replace(/^([^@+])@.*$/, '$1');
-	}
-
-	const dependencyPath = path.join(prefix, name);
-	const children: YarnDependency[] = [];
-
-	for (const child of (tree.children || [])) {
-		const dep = asYarnDependency(path.join(prefix, name, 'node_modules'), child, prune);
-
-		if (dep) {
-			children.push(dep);
+	let dependencyPath;
+	let newPrefix = prefix;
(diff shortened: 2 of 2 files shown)


joaomoreno:
I'm also not too excited to simply duplicate npm's internal behavior. Any chance we can simply depend on `npm` and use its API? https://github.com/microsoft/vscode-vsce/issues/246

nlunets:
Unfortunately not really, running npm-ls on a yarn workspace installed package just results in a lot of errors because the package are not where they are expected.

Furthermore we cannot even rely on the basic node resolution capabilities (require.resolve and require.resolve.paths).
The first one expect a resolveable package (ie: with a clear main script) which is not a guarantee (see https://github.com/babel/babel/issues/5509) 
The second one works based on the current file location (ie the npm.js file) which could be located anywhere in the dependency tree.

From the issue you linked it seems that even npm ls is problematic anyway...
In the end what you want to bundle is not related to who installed it or how they installed it, even npm or yarn can install packages in many different ways and your local setup might affect that.
Following the node resolve logic is (in my opinion) a safer approach as you would now depend on only one logic, which has been stable for years and which works regardless of which package manager installed the dependencies or how they did it :/

nlunets:
@joaomoreno any chance for you to have a deeper review on this ?

joaomoreno:
@nlunets Looking into this. Can you give me a concrete example of an extension which would depend on this?

ullasholla:
@joaomoreno, here's a example repo you can test against: https://github.com/ullasholla/yarn-workspace-test

`yarn` hoists dependencies to the workspace root and hence just following symlinks will not help.

ullasholla:
@joaomoreno, have you had a chance to look at this? Cheers.

bestander:
Thanks @nlunets for sending the patch.
Until the PR is accepted I've published the patch to npm https://www.npmjs.com/package/vsce-yarn-patch

bestander:
And I take some responsibility that yarn ls command sucks in workspaces mode :(
It just slipped through the cracks.
Now it's a bit too late with Yarn V2 being imminent.

@joaomoreno, any chance this can be accepted?
The fix does work on workspaces repos.

src/package.ts
```diff
@@ -837,7 +837,14 @@ export function collect(manifest: Manifest, options: IPackageOptions = {}): Prom
 	const processors = createDefaultProcessors(manifest, options);
 
 	return collectFiles(cwd, useYarn, packagedDependencies).then(fileNames => {
-		const files = fileNames.map(f => ({ path: `extension/${f}`, localPath: path.join(cwd, f) }));
+		const files = fileNames.map(f => {
```
joaomoreno: I'd rather we change the return values of `collectFiles` to return absolute paths instead of hardcoding yarn workspaces knowledge down here.

bestander:
>> And I take some(ok, all) responsibility that yarn ls command sucks in workspaces mode :(
It just slipped through the cracks.
Now it's a bit too late with Yarn V2 being imminent.

> What do you mean?

Never mind, I implemented the workspaces in yarn and did not realize that list command got broken.
It's worth making a fix in Yarn V2 https://github.com/yarnpkg/berry though when it's released

bestander:
I was a bit too eager on this PR, turns out it goes into infinite loop in some cases. 
It needs to maintain a set of visited folders when  looping through node_modules.

bestander:
@nlunets, I've forked your PR and pushed this fix to the infinite loop https://github.com/bestander/vscode-vsce/commit/44d3c26bd055a46a3b7ff49de214f987a3aeb8ad

Let us know if you want to proceed with the PR feedback

nlunets:
@bestander thanks for your involvement ! 
Since you probably have deeper knowledge of how yarn does its magic it might indeed be better if you take care of the PR feedback !
Thanks !

bestander:
Thanks, I'll resubmit the PR on weekend

bestander:
Apologies, everyone.
I was distracted by other work and did not finish cleaning up the PR.
I plan to follow up but can't commit to a specific date yet.

alexgorbatchev:
Ping! Would love to see this merged soon! Thanks for making it happen!

joaomoreno:
Closing since the feedback wasn't addressed and it seems @bestander will simply create a new one.

Issue #391, opened 2019-10-11
vsce package throws error without message


tzfrs:
Hi,

I wanted to build my first VS Code extension, so I followed the tutorial on https://code.visualstudio.com/api/get-started/your-first-extension

So I have a code base containing the src for the hello world extension. 

I now wanted to package this into a `.vsix` file so I can install it on VS Code. So I navigate into my folder (same where package.json is) and execute `vsce package`

This is the output

```
Executing prepublish script 'npm run vscode:prepublish'...

> lorem-ipsum-generator@0.0.1 vscode:prepublish C:\Users\username\dev\lorem-ipsum-generator
> npm run compile


> lorem-ipsum-generator@0.0.1 compile C:\Users\username\dev\lorem-ipsum-generator
> tsc -p ./

 ERROR  Make sure to edit the README.md file before you publish your extension.
```

Here u got the same cmd as screenshot

![image](https://user-images.githubusercontent.com/3509557/66631771-3fc03c80-ec07-11e9-8dca-698cab688e4c.png)

As you can see on the last line, there's an ERROR output, but there's no message. If I run the commands `npm run vscode:prepublish` and `tsc -p ./` independently they don't throw any errors, so I think on this end everything works fine.

Also, the response of `vsce package ls` is

```
Executing prepublish script 'npm run vscode:prepublish'...

> lorem-ipsum-generator@0.0.1 vscode:prepublish C:\Users\username\dev\lorem-ipsum-generator
> npm run compile


> lorem-ipsum-generator@0.0.1 compile C:\Users\username\dev\lorem-ipsum-generator
> tsc -p ./

CHANGELOG.md
out/extension.js
package.json
README.md
```

As you can see, no error in there. It only comes when doing `vsce package`

Versions I'm using:

* vsce: 1.67.1
* npm: 6.9.0
* tsc: 3.6.4

Is there any way I can get an error with a message or log or sth. like that?

joaomoreno:
The message is there, it reads:

```
Make sure to edit the README.md file before you publish your extension.
```

This is there to prevent publishing an extension with the default template README.md file. Edit its contents to describe your extension.

tzfrs:
Oh wow this is embarassing haha. I thought this was more like a warning and not an error. 🤦‍♂ Changing the contents indeed was the fix.

tzfrs:
[...]

PR #392, opened 2019-10-11
Improve error message when packaging an extension with an unchanged README.md

tzfrs:
"Resolves" #391

I put "Resolves" into quotes since it's not really a bug, I just think this improves the experience. The case that a readme is unchanged was also never tested, so I added a test as well. 

@joaomoreno maybe you want to have a look at this and add your 2 cents? :) thank you.


joaomoreno:
Awesome, thanks! :beers:

Issue #389, opened 2019-10-02
Add github.com as badge provider


andreujuanc:
Now with Github Actions having badges, we've got a new provider to add to the whitelist.
I'm removing my badges for now.

joaomoreno:
Since https://github.com/microsoft/vscode-vsce/commit/84f537a4fac4b26b37a3cbd0f3644f0f517a04ec there is no more whitelist. Please update to the latest vsce.

andreujuanc:
@joaomoreno  Well, when i had GitHub badges it stated that it was an invalid badge.

I will try again next week and create another issue if it occurs again.

Cheers

andreujuanc:
@joaomoreno  
https://code.visualstudio.com/api/working-with-extensions/publishing-extension#publishing-extensions
https://code.visualstudio.com/api/references/extension-manifest#approved-badges

All these doc articles does not reflect the fact that the badges are not whitelisted anymore.

And this part of the doc , actually states the problem I have now:
`Images in README.md and CHANGELOG.md may not be SVGs unless they are from trusted badge providers.`

joaomoreno:
@andreujuanc Thanks for the ping.

We've actually just had a discussion that this list should be brought back: https://github.com/microsoft/vscode-vsce/issues/395

I will fix it asap and release a new VSCE version.

joaomoreno:
@andreujuanc Can you show me an example badge from GH?

andreujuanc:
Sorry for the delay @joaomoreno it was a busy day.
Here is the badge I was trying to use:

`[![Actions Status](https://github.com/DigitalMolecules/vscode-servicebus-explorer/workflows/Build/badge.svg)](https://github.com/DigitalMolecules/vscode-servicebus-explorer/actions)`

Thanks for reopening this, it'd be really cool to show the badge :)

joaomoreno:
We can't add that source as a trusted badge provider, since this would enable anyone to simply use any badge by simply uploading it to Github.

joaomoreno:
Hm... I see that's actually the Actions badge. We might need to improve the pattern here, since we currently filter on domain only.

cc @viradhamMS @pkuma-msft

JohnstonCode:
Reference to all the possible badge URLs
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/configuring-a-workflow#adding-a-workflow-status-badge-to-your-repository

felangel:
[...]

PR #396, opened 2019-10-17
Allow badges from Github Workflows

ItsMajestiX:
Fixes https://github.com/microsoft/vscode-vsce/issues/389

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -183,6 +183,10 @@ function isGitHubRepository(repository: string): boolean {
 	return /^https:\/\/github\.com\/|^git@github\.com:/.test(repository || '');
 }
 
+function isGitHubBadge(href: string): boolean {
+	return isGitHubRepository(href) && /[A-Za-z0-9_-]{1,100}\/workflows\/[^<>:;,?"*|/]+\/badge\.svg$/.test(href || '');
+}
+
 class ManifestProcessor extends BaseProcessor {
 
 	constructor(manifest: Manifest) {
@@ -449,7 +453,7 @@ export class MarkdownProcessor extends BaseProcessor {
 				throw new Error(`Images in ${this.name} must come from an HTTPS source: ${src}`);
 			}
 
-			if (/\.svg$/i.test(srcUrl.pathname) && !isHostTrusted(srcUrl.host)) {
+			if (/\.svg$/i.test(srcUrl.pathname) && (!isHostTrusted(srcUrl.host) && !isGitHubBadge(srcUrl.href))) {
 				throw new Error(`SVGs are restricted in ${this.name}; please use other file image formats, such as PNG: ${src}`);
 			}
 		});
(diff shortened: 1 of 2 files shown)


joaomoreno:
@ItsMajestiX Well, what I meant in https://github.com/microsoft/vscode-vsce/issues/389 is that we need to coordinate with the Marketplace. They need to support such filtering mechanism first. Only then we can merge this in VSCE. I'll try to get this discussion started with them.

ItsMajestiX:
@joaomoreno That's fine. I'll leave the PR open until you can get this resolved with the Marketplace team.

zawataki:
@joaomoreno Do you have any progress about the discussion with the Marketplace?

src/package.ts
```diff
@@ -183,6 +183,10 @@ function isGitHubRepository(repository: string): boolean {
 	return /^https:\/\/github\.com\/|^git@github\.com:/.test(repository || '');
 }
 
+function isGitHubBadge(href: string): boolean {
+	return isGitHubRepository(href) && /[A-Za-z0-9_-]{1,100}\/workflows\/[^<>:;,?"*|/]+\/badge\.svg$/.test(href || '');
```
mayankmaheshwari18, line 187: I guess this regex will not allow this url: https://github.com/mayankmaheshwari18/demoRepo/workflows/CI/badge.svg?event=create **which is a valid Svg Url**.

joaomoreno: I suggest instead to go with something much simpler:

```
/^https:\/\/github.com\/[^/]+\/[^/]+\/workflows\/.*badge.svg/
```

Here's a playground to see it in action: https://regexr.com/53238

mayankmaheshwari18: Couple of things:
1) Can we also add \??(.)*$ at the end of Regex so that we take care of the query parameters. 
    Though the current Regex might match it but maybe for any corner cases?
2) Do we also need to take care of https://git@github.com/

Ex: https://git@github.com/mayankmaheshwari18/demoRepo/workflows/CI/badge.svg?event=create

joaomoreno: Adding `??(.)*$` won't matter much, since we're not really interested in looking all the way to the end of the string, right? As long as there's a `badge\.svg` after `\/workflows\/` we should be more than fine. What do you think?

I would also not worry about user names.

mayankmaheshwari18: I thought adding them would make our validations more strict.
Same for the username and repo name. 
But We can go ahead with /^https:\/\/github\.com\/[^/]+\/[^/]+\/workflows\/.*badge\.svg/ 

Any thoughts on https://git@github.com
because this too is a valid url: https://git@github.com/mayankmaheshwari18/demoRepo/workflows/CI/badge.svg?event=create

joaomoreno: I mean `git@` when I referred to user names. I don't really think we need to worry about them, since the official GitHub recommendations for the badges omits them.

mayankmaheshwari18: Oh ok!
I will go ahead with this Regex.
/^https://github.com/[^/]+/[^/]+/workflows/.*badge.svg/
Please let me know if you make some changes in Regex from VS Code side so that we align to that to have better user Experience :)

joaomoreno: I just noticed the dots should be escaped. So the final form would be:

```
/^https://github\.com/[^/]+/[^/]+/workflows/.*badge\.svg/
```

Let me know when the changes make the prod Marketplace, so we update vsce then. Thanks! 🙏

mayankmaheshwari18: Sure! :)

Issue #403, opened 2019-11-29
Validate extensions having string value for extension kind while publishing


sandy081:
We changed  `extensionKind` property in `package.json` to be an array instead of string. Please see https://github.com/microsoft/vscode/issues/85036 and https://github.com/microsoft/vscode/issues/85819 for more details.

Hence we would like to have a check to validate extensions having string value and warn them while publishing.

This is not needed immediately, we want to give extension authors a month or two to adopt and then we can add this check.

PR #404, opened 2019-11-30
Warns the extension authors of `extensionKind` type change

joaomoreno:
Fixes https://github.com/microsoft/vscode-vsce/issues/403

@sandy081:

- This currently only emits a warning, doesn't prevent publishing... should it?
- I believe we need a better user facing message, maybe with a better link describing the adoption and the purpose of that property.

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -42,6 +42,7 @@ export interface Manifest {
 	_testing?: string;
 	enableProposedApi?: boolean;
 	qna?: 'marketplace' | string | false;
+	extensionKind?: string[];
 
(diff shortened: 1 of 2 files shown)


sandy081:
Here is the link to the documentation - https://code.visualstudio.com/api/advanced-topics/remote-extensions#incorrect-execution-location

May be we should show this link.

Warning is good enough I think.

joaomoreno:
Created https://aka.ms/vscode/api/incorrect-execution-location for it.

Left the warning.

joaomoreno:
Coming out in 1.70.0

PR #405, opened 2019-11-30
fixes #400 allow force to unpublish an extension

### LINKED ISSUE #400: Force unpublish extension from marketplace
Hi, I wonder are there any way to force unpublish extension from marketplace instead of interactively asking me `This will FOREVER delete 'XX.XXX'! Are you sure? [y/N]`

The scenario is that I will publish test version extension on marketplace with id "dilin.my-test-extension". The extension is pushed to marketplace in CI/CD. Since the extension is just a test version extension, it's version will always be the same for the same release of main extension. If we fix some issues and would like to publish the test extension to marketplace, considering the version does not change, we need to unpublish the former one from marketplace and publish the new one.

I would like to have a script like `vsce unpublish -p <my-PAT> dilin.my-test-extension` in CI/CD script. The problem is that this command is interactive and will ask me to confirm `Yes`. Are there any way to force unpublish extension from marketplace? A `-f` option for example.
THX!

Grommers00:
fixes #400

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -93,7 +93,8 @@ module.exports = function (argv: string[]): void {
 		.command('unpublish [<extensionid>]')
 		.description('Unpublishes an extension. Example extension id: microsoft.csharp.')
 		.option('-p, --pat <token>', 'Personal Access Token')
-		.action((id, { pat }) => main(unpublish({ id, pat })));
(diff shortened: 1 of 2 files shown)


joaomoreno:
Thanks! :fireworks:

PR #409, opened 2019-12-19
Modifying messages to match the current situation

wraith13:
I know it was a few seconds before. But now it is a few minutes.


joaomoreno:
Ahah yeah. Good change. Thanks!

PR #412, opened 2019-12-29
Log level for npm list command changed to error to increase compatibility

xontab:
* Changed log level to error instead of warning for `npm list` command


joaomoreno:
Thanks! :beers:

Issue #415, opened 2020-01-20
Four number version string with semver


sguillia:
Hello

---

In my package.json

```
    "version": "0.0.1-1"
```

Output of _vsce publish_

```
 ERROR  The version string '0.0.1-1' doesn't conform to the requirements for a version. It must be one to four numbers in the range 0 to 2147483647, with each number seperated by a period. It must contain at least one non-zero number.

```

---

In my package.json

```
    "version": "0.0.1.1"
```

Output of _vsce publish_

```
 ERROR  Invalid extension version '0.0.1.1'
```

---

Does vsce support any four-number version string?

I believe, if I am not mistaken, that the former error message is wrong.

Best,
Simon

joaomoreno:
Semver doesn't not support four numbers: https://semver.org/

sguillia:
@joaomoreno Semver does support four numbers, the fourth one is called prerelease.

If it would not, vsce erroneously says that the version string `must be one to four numbers`.

Please update the error message if you do not support prereleases.

joaomoreno:
Oh I see. Actually, the first error message seems to be coming from the server. I'll update VSCE to catch that error earlier and let the user know what is supported.

sguillia:
Many thanks

sean-mcmanus:
@joaomoreno This change breaks our C/C++ extension's vsix publishing. We create vsix's with versions like 0.26.3-insiders4 (https://github.com/microsoft/vscode-cpptools/releases/tag/0.26.3-insiders4) and 0.27.0-multiroot2 for internal testing. Can you re-enable our scenario to work?

WardenGnaw:
> ERROR  The version string '0.0.1-1' doesn't conform to the requirements for a version. It must be one to **four** numbers in the range 0 to 2147483647, with each number seperated by a period. It must contain at least one non-zero number.
The four in this error message should be **three**. 

Since semver expects <major>.<minor>.<patch>[-<prerelease>]

The documentation on prerelease is:
[...]

PR #420, opened 2020-02-02
revert e4a8df59

eine:
Ref #415

> Is it possible that the server supports pre-releases starting with non-digit characters only? Should that be the case, I would ask to revert [e4a8df5](https://github.com/microsoft/vscode-vsce/commit/e4a8df59d34e1f6a83776256067cf11bff943f1f) ASAP.

**EDIT**

Actually, according to the CI job, this seems to be fixed upstream.


joaomoreno:
I have reverted the commit which disabled prerelase versions, but the Marketplace will keep not supporting them.

PR #424, opened 2020-02-20
Add --noGitHubIssueLinking to stop issue link expansion in package step

rjmholt:
Adds an option to `vsce package` to prevent any expansion of GitHub issue references.

**Note**: This is a workaround for https://github.com/microsoft/vscode-vsce/issues/398, but is not a fix for that issue.


joaomoreno:
Thanks!

Issue #430, opened 2020-03-11
Can we change the logging method for prepublish?


bobbrow:
This line here is causing errors to suddenly appear in our Build Pipeline:
https://github.com/microsoft/vscode-vsce/blob/66a41f19ab006a5092373f8cd90b873bbfae5bbd/src/package.ts#L930

![image](https://user-images.githubusercontent.com/12818240/76467449-e4a59800-63a6-11ea-83dc-aa9532dc1ced.png)

This code seems to have been this way for a while, so maybe it's a new behavior in Azure Dev Ops.  Would it be possible to change it to `console.log` instead?  It doesn't seem like "warning" is the right classification for this. I can send a PR if you like.

joaomoreno:
Yes, PR accepted.

PR #434, opened 2020-03-31
Change a console.warn to console.log for an informational message

bobbrow:
Azure DevOps is reporting this warning as an error.  Since this message is not really a warning, just a notification of which package tool is being used, it can be changed to `console.log`


bobbrow:
#430

bobbrow:
Thanks! 😃

PR #436, opened 2020-04-01
Migrate to a better matching engine

### LINKED ISSUE #428: Migrate to leven to find the closest match for an unknown command
As of now [didyoumean.js](https://github.com/dcporter/didyoumean.js/) matching engine is used to suggest matching commands if the user mistypes. But it is not longer actively maintained. Instead I would suggest switching over to [leven](https://github.com/sindresorhus/leven) which is the fastest JS implementation of the Levenshtein distance algorithm.

`leven` returns the distance between 2 strings which can be made use of to find the exact match.

jamesgeorge007:
closes #428

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
   "devDependencies": {
     "@types/cheerio": "^0.22.1",
     "@types/denodeify": "^1.2.31",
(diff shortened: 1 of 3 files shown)


joaomoreno:
Looks great, thanks!

PR #447, opened 2020-05-11
Support injecting PAT from an environmental variable

owenfarrell:
This change adds a default value to the `-p` (`--pat`) option for the publish command based on an environmental variable.

Making this change creates an implicit default value for providing a personal access token, making it easier to trigger VSCE releases from CI platforms where global installation of modules is not possible (i.e. publish a VSCE extension via an `npm run...` command).


joaomoreno:
I love it, thanks! 🔥

Issue #432, opened 2020-03-23
Created VSIX via `vsce package --yarn` does not include not-hoisted deep dependency


yhatt:
Think about a simple package like this, developed with yarn 1.22.4:

```json
{
  "name": "vsce-yarn-lacked-dependency",
  "description": "example",
  "version": "0.0.1",
  "publisher": "yhatt",
  "repository": "https://github.com/yhatt/vsce-yarn-lacked-dependency",
  "engines": {
    "vscode": "^1.40.0"
  },
  "activationEvents": [
    "*"
  ],
  "main": "./extension.js",
  "devDependencies": {
    "vsce": "^1.74.0"
  },
  "dependencies": {
    "markdown-it": "^10.0.0"
  }
}
```

```javascript
// extension.js
const md = require("markdown-it")();

module.exports = {
  activate: () => console.log(md.render("Hello, world!"))
};
```

The creation of extension through `yarn vsce package --yarn` will success, but will fail to activate in VS Code.

```
abstractExtensionService.ts:396 Activating extension 'yhatt.vsce-yarn-lacked-dependency' failed: Cannot find module 'entities/lib/maps/entities.json'
Require stack:
- /Users/yuki.hattori/.vscode/extensions/yhatt.vsce-yarn-lacked-dependency-0.0.1/node_modules/markdown-it/lib/common/entities.js
- /Users/yuki.hattori/.vscode/extensions/yhatt.vsce-yarn-lacked-dependency-0.0.1/node_modules/markdown-it/lib/common/utils.js
- /Users/yuki.hattori/.vscode/extensions/yhatt.vsce-yarn-lacked-dependency-0.0.1/node_modules/markdown-it/lib/index.js
- /Users/yuki.hattori/.vscode/extensions/yhatt.vsce-yarn-lacked-dependency-0.0.1/node_modules/markdown-it/index.js
- /Users/yuki.hattori/.vscode/extensions/yhatt.vsce-yarn-lacked-dependency-0.0.1/extension.js
- /Applications/Visual Studio Code.app/Contents/Resources/app/out/vs/loader.js
- /Applications/Visual Studio Code.app/Contents/Resources/app/out/bootstrap-amd.js
- /Applications/Visual Studio Code.app/Contents/Resources/app/out/bootstrap-fork.js.
```

It would work correctly if created through npm with `yarn vsce package`.

[...]

PR #461, opened 2020-06-26
Update launch.json to be useful for extension debugging

viktomas:
When I worked on debugging https://github.com/microsoft/vscode-vsce/issues/432#issuecomment-650148223, I needed to make the extension work locally. This configuration change was very useful for running a local development version to package an extension.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -8,13 +8,14 @@
 			"type": "node",
 			"request": "launch",
 			"name": "Launch Program",
+			// "cwd": "<absolute path to your extension>",
 			"program": "${workspaceFolder}/out/vsce",
 			"args": [
 				"--version"
+				// "ls", "package", "publish"
 			],
 			"sourceMaps": true,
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -8,13 +8,13 @@
...
 			"name": "Launch Program",
+			// "cwd": "<absolute path to your extension>",
 			"program": "${workspaceFolder}/out/vsce",
 			"args": [
-				"--version"
+				"ls"
 			],
 			"sourceMaps": true,
-			"outputCapture": "std",
-			"preLaunchTask": "compile"
```
viktomas, line 17: This task isn't configured.

joaomoreno:
Isn't `ls` gonna fail if you don't provide a `cwd`? How is this better than the current option?

viktomas:
> Isn't ls gonna fail if you don't provide a cwd?

That's a good point. I got carried away with my usecase. Please see updated change @joaomoreno

PR #472, opened 2020-07-31
Prevent mailto links from being joined with prefix

### LINKED ISSUE #228: mailto links in README.md don't work with "vsce package"
Repro steps:
1. Add mailto link in README.md, e.g. \[Email me!](mailto:me@website.com?Subject=Repro).
2. Run "vsce package"
3. I get the following error:
Error: Couldn't detect the repository where this extension is published. The link 'mailto:me@website.com?Subject=Repro' will be broken in README.md. Please provide the repository URL in package.json or use the --baseContentUrl and --baseImagesUrl options.

If I run "vsce package --baseContentUrl" it ends up adding "true/" to the link like so:
\[Email me!](true/mailto:me@website.com?Subject=Repro)

I would expect vsce to ignore mailto links such that they actually work when the package is installed and the user views the readme.

JeffreyCA:
Resolves #228

Updated some tests and they all pass.

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -390,7 +390,11 @@ export class MarkdownProcessor extends BaseProcessor {
 		}
 
 		const markdownPathRegex = /(!?)\[([^\]\[]*|!\[[^\]\[]*]\([^\)]+\))\]\(([^\)]+)\)/g;
-		const urlReplace = (_, isImage, title, link) => {
+		const urlReplace = (_, isImage, title, link: string) => {
+			if (/^mailto:/i.test(link)) {
+				return `${isImage}[${title}](${link})`;
+			}
+
 			const isLinkRelative = !/^\w+:\/\//.test(link) && link[0] !== '#';
 
 			if (!this.baseContentUrl && !this.baseImagesUrl) {
--- a/src/test/fixtures/readme/readme.expected.md
+++ b/src/test/fixtures/readme/readme.expected.md
@@ -31,6 +31,7 @@ The `spellMD.json` config file is watched so you can add more ignores or change 
 
 [mono](https://github.com/username/repository/blob/master/monkey)
 [not](http://shouldnottouchthis/)
+[Email me](mailto:example@example.com)
 
 # Install
 This extension is published in the VS Code Gallery.  So simply hit 'F1' and type 'ext inst' from there select `SpellMD` and follow instructions.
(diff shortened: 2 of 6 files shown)


src/package.ts
```diff
@@ -390,13 +390,14 @@ export class MarkdownProcessor extends BaseProcessor {
 		}
 
 		const markdownPathRegex = /(!?)\[([^\]\[]*|!\[[^\]\[]*]\([^\)]+\))\]\(([^\)]+)\)/g;
-		const urlReplace = (_, isImage, title, link) => {
+		const urlReplace = (_, isImage, title, link: string) => {
 			const isLinkRelative = !/^\w+:\/\//.test(link) && link[0] !== '#';
+			const isMailTo = link.toLowerCase().startsWith('mailto:');
```
joaomoreno, line 395:
```suggestion
			if (/^mailto:/i.test(link)) {
				return link;
			}
			
			const isLinkRelative = !/^\w+:\/\//.test(link) && link[0] !== '#';
```

src/package.ts
```diff
@@ -390,13 +390,14 @@ export class MarkdownProcessor extends BaseProcessor {
...
 
 			if (!this.baseContentUrl && !this.baseImagesUrl) {
 				const asset = isImage ? 'image' : 'link';
 
-				if (isLinkRelative) {
+				if (isLinkRelative && !isMailTo) {
```
joaomoreno, line 400:
```suggestion
				if (isLinkRelative) {
```

src/package.ts
```diff
@@ -408,7 +409,8 @@ export class MarkdownProcessor extends BaseProcessor {
 				return `${isImage}[${title}](${link})`;
 			}
 
-			return `${isImage}[${title}](${urljoin(prefix, link)})`;
+			const finalLink = isMailTo ? link : urljoin(prefix, link);
+			return `${isImage}[${title}](${finalLink})`;
```
joaomoreno, line 413:
```suggestion
			return `${isImage}[${title}](${urljoin(prefix, link)})`;
```

src/package.ts
JeffreyCA: Thanks for the suggestions. I believe this should return `${isImage}[${title}](${link})` instead.

joaomoreno:
Thanks!

PR #474, opened 2020-08-09
Add githubBranch flag to control the branch for GitHub relative links

### LINKED ISSUE #468: Extension marketplace readme renderer assumes you're using branch called master
- VSCode Version: 1.47.2
- OS Version: Windows 10 1903

## Steps to Reproduce:

1. Create a new extension with links to images in its `README.md` file

For example,

```md
![some image](./images/some-image.png)
```

2. Publish the extension

## Expected

Images work

## Actual

Images appear to link to i.e. `https://github.com/user/extension/raw/master/images/some-image.png`

If, for example, your default branch is called `main` then this means that all the images are broken.

This also assumes that the extension lives at the root of the repo; this will break in the event that an extension is in a subfolder.

Does this issue occur when all extensions are disabled?: Yes

## Workaround

Directly link to the images in the GitHub repo. Don't have relative image paths in your extension's `README.md`.

JeffreyCA:
Fixes #468 

Currently if `repository` in `package.json` is set to a GitHub URL, all the relative links in `README.md` get converted to absolute links but they always refer to the `master` branch. This change adds a flag `githubBranch` which allows the user to override this assumption without having to use to override `baseImagesUrl` and `baseContentUrl`.

The `baseImageUrl` and `baseContentUrl` options take precedence over `githubBranch`, so doing something like: `vsce --githubBranch main --baseImagesUrl http://blah` will only change non-image relative links to stem from the `main`.

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -59,6 +59,7 @@ export interface IAsset {
 export interface IPackageOptions {
 	cwd?: string;
 	packagePath?: string;
+	githubBranch?: string;
 	baseContentUrl?: string;
 	baseImagesUrl?: string;
 	useYarn?: boolean;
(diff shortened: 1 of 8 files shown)


src/package.ts
```diff
@@ -514,10 +515,11 @@ export class MarkdownProcessor extends BaseProcessor {
 
 		const account = match[1];
 		const repositoryName = match[2].replace(/\.git$/i, '');
+		const branchName = githubBranch ? githubBranch : 'master';
```
joaomoreno, line 518:
```suggestion
		const branchName = githubBranch ?? 'master';
```

JeffreyCA: The current TypeScript version is `3.4.3` which doesn't support nullish coalescing. Is it okay to update the packages? I can revert the changes otherwise.

joaomoreno:
Thanks! 🎆

PR #487, opened 2020-09-16
Don't check for publisher in vsce package

### LINKED ISSUE #328: vsce package should not check for publisher
Repro:

- Clone a sample from https://github.com/Microsoft/vscode-extension-samples
- Run `vsce package`
- `vsce` should not check for `publisher` for generating the VSIX

vojtechkral:
Fix #328

This is quite an annoying bug when one wants to publish a private / organization-internal extension to co-workers...

I fixed it in a really simple way, not sure it this is good enough...

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -14,7 +14,7 @@ import * as cheerio from 'cheerio';
 import * as url from 'url';
 import { lookup } from 'mime';
 import * as urljoin from 'url-join';
-import { validatePublisher, validateExtensionName, validateVersion, validateEngineCompatibility, validateVSCodeTypesCompatibility } from './validation';
+import { validateExtensionName, validateVersion, validateEngineCompatibility, validateVSCodeTypesCompatibility } from './validation';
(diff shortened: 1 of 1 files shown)


vojtechkral:
> I fixed it in a really simple way, not sure it this is good enough...

In fact, it's not. Just removing the function causes VS Code to error out with `Cannot read property 'toLocaleLowerCase' of undefined`.

Not sure what to do about that...

Supitto:
@vojtechkral I am build a independent packager and stumbled upon this. Apparently the function installFromVSIX https://github.com/microsoft/vscode/blob/a416c1534bb9159c7ad9d88f19d70019ded16a6a/src/vs/workbench/contrib/extensions/browser/extensionsWorkbenchService.ts#L1032) uses a field 'publisher' from the extension. If you add the field 'publisher' into the package.json like this

`{
  "publisher": "test",
  "name": "something",
`

vscode will recognize this as the publisher and properly register the extension.

PR #458, opened 2020-06-22
feat: Support for yarn workspaces

sebas2day:
Fixes #300

This PR is inspired by #379.

I have included a small yarn workspaces fixture in order to test whether its collecting the files in the right way.

PR #492, opened 2020-09-24
Adds detection for yarn, and '--no-yarn' to suppress detection

### LINKED ISSUE #480: Automatically enable `--yarn` if `yarn.lock` is found
How about that guys?

rbuckton:
This adds automatic detection of `yarn` based on the presence of either a `yarn.lock` or `.yarnrc` file in the current directory. Passing `--yarn` or `--no-yarn` disables automatic detection.

Fixes #480

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -61,7 +61,8 @@ module.exports = function (argv: string[]): void {
 	program
 		.command('ls')
 		.description('Lists all the files that will be published')
-		.option('--yarn', 'Use yarn instead of npm')
+		.option('--yarn', 'Use yarn instead of npm (default inferred from presense of yarn.lock or .yarnrc)')
+		.option('--no-yarn', 'Use npm instead of yarn (default inferred from lack of yarn.lock or .yarnrc)')
 		.option(
 			'--packagedDependencies <path>',
 			'Select packages that should be published only (includes dependencies)',
@@ -83,7 +84,8 @@ module.exports = function (argv: string[]): void {
 		)
 		.option('--baseContentUrl [url]', 'Prepend all relative links in README.md with this url.')
 		.option('--baseImagesUrl [url]', 'Prepend all relative image links in README.md with this url.')
-		.option('--yarn', 'Use yarn instead of npm')
+		.option('--yarn', 'Use yarn instead of npm (default inferred from presense of yarn.lock or .yarnrc)')
+		.option('--no-yarn', 'Use npm instead of yarn (default inferred from lack of yarn.lock or .yarnrc)')
 		.option('--ignoreFile [path]', 'Indicate alternative .vscodeignore')
 		.option('--noGitHubIssueLinking', 'Prevent automatic expansion of GitHub-style issue syntax into links')
 		.option(
@@ -117,7 +119,8 @@ module.exports = function (argv: string[]): void {
 		)
 		.option('--baseContentUrl [url]', 'Prepend all relative links in README.md with this url.')
 		.option('--baseImagesUrl [url]', 'Prepend all relative image links in README.md with this url.')
-		.option('--yarn', 'Use yarn instead of npm while packing extension files')
+		.option('--yarn', 'Use yarn instead of npm (default inferred from presense of yarn.lock or .yarnrc)')
+		.option('--no-yarn', 'Use npm instead of yarn (default inferred from lack of yarn.lock or .yarnrc)')
 		.option('--noVerify')
 		.option('--ignoreFile [path]', 'Indicate alternative .vscodeignore')
 		.option(
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -181,7 +181,10 @@ async function getYarnProductionDependencies(cwd: string, packagedDependencies?:
 async function getYarnDependencies(cwd: string, packagedDependencies?: string[]): Promise<string[]> {
 	const result: string[] = [cwd];
 
-	if (await new Promise(c => fs.exists(path.join(cwd, 'yarn.lock'), c))) {
+	// Using 'existsSync' instead of 'exists' since 'exists' is deprecated. While 'fs.existsSync'
+	// is a blocking IO operation, this won't affect the CLI as every code path that calls
+	// 'getYarnDependencies' essentially blocks the CLI anyways while waiting for `exists` to complete.
+	if (fs.existsSync(path.join(cwd, 'yarn.lock'))) {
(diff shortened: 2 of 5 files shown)


src/package.ts
```diff
@@ -1124,14 +1124,14 @@ function getDefaultPackageName(manifest: Manifest): string {
...
-async function prepublish(cwd: string, manifest: Manifest, useYarn: boolean = false): Promise<void> {
+async function prepublish(cwd: string, manifest: Manifest, useYarn: boolean = detectYarn(cwd)): Promise<void> {
 	if (!manifest.scripts || !manifest.scripts['vscode:prepublish']) {
 		return;
 	}
 
 	console.log(`Executing prepublish script '${useYarn ? 'yarn' : 'npm'} run vscode:prepublish'...`);
 
-	await new Promise((c, e) => {
+	await new Promise<void>((c, e) => {
```
rbuckton, line 1134: In TypeScript 4.1, the value to `resolve()` is no longer optional unless a `Promise` is given the type `Promise<void>` (either explicitly or via a contextual type). Even though vsce currently uses TypeScript 3.4, I added this because it'll be an issue in the future (also, I am running VS Code Insiders with the latest TS 4.1 build and it showed up as a diagnostic).

felipecrs (approved):
LGTM

rbuckton:
This might need to be updated if #458 goes in, as this currently only looks for `yarn.lock` and `.yarnrc` in the current directory, but if you're publishing an from a package subdirectory in a yarn workspace, neither of those two files will be present since the current directory is not the workspace root.

rbuckton:
Build is failing before it starts due to the node version used to run the build:

```
yarn install v1.22.5
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
error husky@4.3.0: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

https://dev.azure.com/vscode/VSCE/_build/results?buildId=53730&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=f8ed7bd8-2a7f-56f6-9385-7fc29a8b5b7b&l=18

rbuckton:
@joaomoreno I bumped the node version in `azure-pipelines.yml` to get the build to pass. I can revert that if you still want to be sure it builds/tests on NodeJS 8, but I can't do much about husky 4.3.0 not supporting node 8.

src/package.ts
```diff
@@ -1157,6 +1157,16 @@ async function getPackagePath(cwd: string, manifest: Manifest, options: IPackage
 	}
 }
 
+function detectYarn(cwd: string) {
+	for (const file of ["yarn.lock", ".yarnrc"]) {
+		if (fs.existsSync(path.join(cwd, file))) {
```
joaomoreno, line 1162: Using `sync` is out of question. We should rewrite this to be fully async.

src/package.ts
```diff
@@ -1157,6 +1157,16 @@ async function getPackagePath(cwd: string, manifest: Manifest, options: IPackage
...
+			console.log(`Detected presense of ${file}. Using 'yarn' instead of 'npm' (to override this pass '--no-yarn' on the command line).`);
```
joaomoreno, line 1163: We have a logging util in `util.ts`, let's use that instead.

src/npm.ts
```diff
@@ -181,7 +181,10 @@ async function getYarnProductionDependencies(cwd: string, packagedDependencies?:
 async function getYarnDependencies(cwd: string, packagedDependencies?: string[]): Promise<string[]> {
 	const result: string[] = [cwd];
 
-	if (await new Promise(c => fs.exists(path.join(cwd, 'yarn.lock'), c))) {
+	// Using 'existsSync' instead of 'exists' since 'exists' is deprecated. While 'fs.existsSync'
+	// is a blocking IO operation, this won't affect the CLI as every code path that calls
+	// 'getYarnDependencies' essentially blocks the CLI anyways while waiting for `exists` to complete.
+	if (fs.existsSync(path.join(cwd, 'yarn.lock'))) {
```
joaomoreno, line 187: We can't use sync because there are other operations which might run in the background, ie checking for updates.

This needs to stay as async, and `stat` should be used.

joaomoreno (changes requested):

src/main.ts
```diff
@@ -61,7 +61,8 @@ module.exports = function (argv: string[]): void {
 	program
 		.command('ls')
 		.description('Lists all the files that will be published')
-		.option('--yarn', 'Use yarn instead of npm')
+		.option('--yarn', 'Use yarn instead of npm (default inferred from presense of yarn.lock or .yarnrc)')
+		.option('--no-yarn', 'Use npm instead of yarn (default inferred from lack of yarn.lock or .yarnrc)')
```
joaomoreno, line 65: Should this perhaps be `--npm`?

rbuckton: The arg parser you're using (commander), has specific handling for `--foo` and `--no-foo`: https://www.npmjs.com/package/commander#other-option-types-negatable-boolean-and-flagvalue. If you make the flag `--npm`, then you have to manually disambiguate and report errors for using `--npm --yarn`

src/package.ts
rbuckton: That's a bit more complex because you can't use `await` in a parameter initializer, and the version of TypeScript you're using doesn't support `??`. I'll look into it though.

src/main.ts
joaomoreno: Oh yeah, cool.

src/package.ts
```diff
@@ -1124,14 +1123,16 @@ function getDefaultPackageName(manifest: Manifest): string {
 	return `${manifest.name}-${manifest.version}.vsix`;
 }
 
-async function prepublish(cwd: string, manifest: Manifest, useYarn: boolean = false): Promise<void> {
+async function prepublish(cwd: string, manifest: Manifest, useYarn?: boolean): Promise<void> {
 	if (!manifest.scripts || !manifest.scripts['vscode:prepublish']) {
 		return;
 	}
 
+	if (useYarn === undefined) useYarn = await detectYarn(cwd);
```
joaomoreno, line 1131:
```suggestion
	if (useYarn === undefined) {
		useYarn = await detectYarn(cwd);
	}
```

joaomoreno (approved):
Thanks!

PR #495, opened 2020-10-03
Fix useYarn always true

### LINKED ISSUE #494: `--yarn` can not work correctly after 1.81.0
If project has `yarn.lock` file, `vsce` will use `yarn` automatically. However, when using `--yarn` flag, it will fall back to `npm`

felipecrs:
Refs https://github.com/tj/commander.js/issues/979.

Fixes #494.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
 		"azure-devops-node-api": "^7.2.0",
 		"chalk": "^2.4.2",
 		"cheerio": "^1.0.0-rc.1",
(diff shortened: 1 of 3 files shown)


felipecrs:
This fixed the issue in my local and the tests keep passing. It would be good if we could add better test cases though.

joaomoreno:
Thanks for the PR @felipecrs. For dependency changes, we need to do the changes ourselves, though. I've pushed a similar commit and will release a new version.

PR #503, opened 2020-10-16
fix: handle unknown args

### LINKED ISSUE #502: handler for unknown arguments throws an error
An unrecognized subcommand results in the following error message:

Screenshot 2020-10-16 at 10 30 16 PM

jamesgeorge007:
closes #502


joaomoreno:
Thanks! 🙏

PR #508, opened 2020-11-18
Ensure package.json is writable in *.vsix

FabianHenneke:
VSC requires the manifest file unpacked from the .vsix archive to be
marked as writable (see [1]). Since generating writable artifacts is
difficult in some build systems (such as Bazel), it would be helpful if
vsce itself could ensure that the file mode of the manifest is correct.

With this commit, the correct file mode on `package.json` is set
automatically by the packaging process via a new `Processor` and yazl's
`mode` argument.

[1]: https://github.com/microsoft/vscode/blob/88144f6d31718288c7a2c6fb741e0646d40804cf/src/vs/platform/extensionManagement/node/extensionsScanner.ts#L144

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -210,7 +210,7 @@ function isHostTrusted(url: url.UrlWithStringQuery): boolean {
 	return TrustedSVGSources.indexOf(url.host.toLowerCase()) > -1 || isGitHubBadge(url.href);
 }
 
-class ManifestProcessor extends BaseProcessor {
+export class ManifestProcessor extends BaseProcessor {
 	constructor(manifest: Manifest) {
 		super(manifest);
 
--- a/src/package.ts
+++ b/src/package.ts
@@ -37,11 +37,13 @@ const MinimatchOptions: minimatch.IOptions = { dot: true };
 
 export interface IInMemoryFile {
 	path: string;
+	mode?: number;
 	readonly contents: Buffer | string;
 }
 
 export interface ILocalFile {
 	path: string;
+	mode?: number;
 	readonly localPath: string;
 }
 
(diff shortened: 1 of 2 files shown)


src/package.ts
```diff
@@ -37,11 +37,13 @@ const MinimatchOptions: minimatch.IOptions = { dot: true };
 
 export interface IInMemoryFile {
 	path: string;
+	explicitMode?: number;
```
joaomoreno, line 40: Let's just call this `mode`.

src/package.ts
```diff
@@ -37,11 +37,13 @@ const MinimatchOptions: minimatch.IOptions = { dot: true };
...
 	readonly contents: Buffer | string;
 }
 
 export interface ILocalFile {
 	path: string;
+	explicitMode?: number;
```
joaomoreno, line 46: Let's just call this `mode`.

joaomoreno (changes requested):
Makes sense. But why not simply implement `onFile` on the `ManifestProcessor` and change the file mode when the `package.json` file comes along? This would be a much smaller change.

FabianHenneke:
Thanks for the suggestion, I have made the required changes.

src/package.ts
```diff
@@ -274,6 +276,32 @@ class ManifestProcessor extends BaseProcessor {
...
+				mode,
+				contents: file.contents,
+			};
+		} else {
+			return {
+				path: file.path,
+				mode,
+				localPath: file.localPath,
+			};
+		}
```
joaomoreno, line 302: Would this work too?

```suggestion
		return { ...file, mode };
```

FabianHenneke: It does, thanks!

joaomoreno (commented):
Looking great! Wanna try adding a test for it? It should be relatively straightforward.

FabianHenneke:
I added a unit test for `MetadataProcessor`. Is that what you intended?

joaomoreno:
Great work, thanks! 🍻

fmeum:
Would it be possible to get an NPM release of 1.83.0?

PR #532, opened 2021-02-03
Add verify-pat command

### LINKED ISSUE #525: New command: verify-pat
Or something like that. Since apparently there is no public API in the VSCode Marketplace for that purpose, it would be handy to have a command which could verify if the PAT is actually usable.

I need this for validating the token as a verification step in the semantic-release-vsce plugin.

https://github.com/raix/semantic-release-vsce/blob/59e9fa4e0ed090388d4a475e7bbbc2d5bdc10df0/lib/verify.js#L17

felipecrs:
Closes #525

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -4,6 +4,7 @@ import { home } from 'osenv';
 import { read, getGalleryAPI, getSecurityRolesAPI, log } from './util';
 import { validatePublisher } from './validation';
 import * as denodeify from 'denodeify';
+import { readManifest } from './package';
 
 const readFile = denodeify<string, string, string>(fs.readFile);
 const writeFile = denodeify<string, string, object, void>(fs.writeFile as any);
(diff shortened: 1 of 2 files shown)


src/store.ts
```diff
@@ -53,14 +54,39 @@ function removePublisherFromStore(store: IStore, publisherName: string): Promise
...
+	if (!pat) {
+		throw new Error('The Personal Access Token is mandatory.');
+	}
+
+	if (!publisherName) {
+		try {
+			publisherName = (await readManifest()).publisher;
+		} catch (error) {
+			throw new Error(
+				'Can not read the publisher name. Either supply it as argument or call from the package root folder. Additional information:\n\n' +
```
joaomoreno, line 67:
```suggestion
				'Can not read the publisher name. Either supply it as argument or run vsce from the extension folder. Additional information:\n\n' +
```

joaomoreno (approved):

joaomoreno:
Thanks!

PR #540, opened 2021-03-04
Use HEAD as default branch rather than master

### LINKED ISSUE #534: Try to detect default GitHub branch
More and more people are changing their default branch from `master` to `main`. It would be great if VSCE could detect this and do right right thing when publishing an extension instead of assuming `master`.

felipecrs:
Fixes #534

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 > _The Visual Studio Code Extension Manager_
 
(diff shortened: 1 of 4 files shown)


joaomoreno:
OK that is really clever, I gotta say I didn't think of this solution. Thanks a lot @felipecrs!

Issue #542, opened 2021-03-09
--noGitHubIssueLinking option doesn't work properly


horihiro:
According to the following code, `this.expandGitHubIssueLinks` is set to `true` when the option is not only specified but also not specified.

https://github.com/microsoft/vscode-vsce/blob/f186af2ab56ff78725bf9285c09b52f64e0988f7/src/main.ts#L104
https://github.com/microsoft/vscode-vsce/blob/f186af2ab56ff78725bf9285c09b52f64e0988f7/src/package.ts#L441-L442

So GitHub issue links in Markdown files are always expanded regardless of the option.

horihiro:
The test for `expandGitHubIssueLinks` option of ReadmeProcessor exists, this test abehavior when `expandGitHubIssueLinks` is `false`.

https://github.com/microsoft/vscode-vsce/blob/f186af2ab56ff78725bf9285c09b52f64e0988f7/src/test/package.test.ts#L1991

But as I commented, `expandGitHubIssueLinks` is always `true`.

PR #543, opened 2021-03-09
set `expandGitHubIssueLinks` to inverse of `noGitHubIssueLinking`

horihiro:
This PR contains handling `noGitHubIssueLinking` option and variable `expandGitHubIssueLinks` properly.

According to `commandar`, `noGitHubIssueLinking` can be `true` or `undefined`, cannot be `false` because the option name doesn't start with `--no-`.
So `expandGitHubIssueLinks` is same.
https://github.com/tj/commander.js#other-option-types-negatable-boolean-and-booleanvalue

On current implementation, the `expandGitHubIssueLinks` is always `true` after this conditional operator if `expandGitHubIssueLinks` cannot be `false`, as reported in #542 
https://github.com/microsoft/vscode-vsce/blob/f186af2ab56ff78725bf9285c09b52f64e0988f7/src/package.ts#L441-L442

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -101,7 +101,7 @@ module.exports = function (argv: string[]): void {
 					baseImagesUrl,
 					useYarn: yarn,
 					ignoreFile,
-					expandGitHubIssueLinks: noGitHubIssueLinking,
+					expandGitHubIssueLinks: !noGitHubIssueLinking,
 					web,
 				})
 			)


joaomoreno:
A better fix is to just adopt the `--no` syntax: https://github.com/microsoft/vscode-vsce/commit/83163ed09e26b2218832f60acbe7aad215c77a02

Thanks though!

PR #555, opened 2021-04-09
Allow version bump in package command

felipecrs:
And add the `--no-git-tag-version` option to disable the creation of version commit and tag when calling `npm version`.

## Plus

- Set prettier as the default formatter for JSON and Typescript files in VS Code and add it to the list of recommended extensions
- Improve `package --out` help message
- Format `main.ts` with Prettier
- Move `versionBump` from `publish.ts` to `package.ts` as it is where it is used first now
- Use `ts-node` with `mocha`
- Upgrade `tmp`
- Revamp the `package.json` scripts by simplifying and renaming some

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -2,7 +2,7 @@
 	// See http://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
 	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
 	// List of extensions which should be recommended for users of this workspace.
-	"recommendations": ["hbenl.vscode-mocha-test-adapter"],
+	"recommendations": ["hbenl.vscode-mocha-test-adapter", "esbenp.prettier-vscode"],
 	// List of extensions recommended by VS Code that should not be recommended for users of this workspace.
 	"unwantedRecommendations": []
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,7 +6,6 @@
 		"out": true
 	},
 	"typescript.tsdk": "./node_modules/typescript/lib",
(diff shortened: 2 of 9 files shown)


joaomoreno:
This looks great. Could we perhaps add some tests?

joaomoreno:
@felipecrs Did you screw up the code with that force push, perhaps? The `versionBump` function is now duplicated.

felipecrs:
@joaomoreno it could be, sorry. Actually I'm still missing the test cases which I'm going to add.

felipecrs:
@joaomoreno Ok I just made the required changes. Some caveats:

- I had to update the "tmp" module, which from what I can say did not break anything
- I changed a bit how mocha works by making it use `ts-node` instead of relying in the test cases previously built. Some advantages in doing that: 
    
    3H9u5wPY4h


    None of this was possible before. Also, this drops the dependency on `tsc` when running tests.

felipecrs:
I think I found a loose end though: 

- `pre` bump strings such as `prepatch` are not allowed, nonetheless, versions previously set with prereleases are allowed and passes the versionValidation, such as `1.0.0-beta.1`

So, should we remove the check for `pre` version bumps and let it go?

joaomoreno:
> So, should we remove the check for `pre` version bumps and let it go?

No, I believe the Marketplace won't take those versions, so better keep it this way.

Cool tests and refactorings! Thanks! I have further cleaned this up. Unsure why you downgraded `ts-node`, though, latest seems to work just fine, so I have upgraded it to latest.

felipecrs:
> Unsure why you downgraded ts-node, though, latest seems to work just fine, so I have upgraded it to latest.

It was because the CI was running under Node 10, which ts-node@10 have dropped support.

PR #558, opened 2021-04-13
Add support for GitLab.com

firelizzard18:
This MR:

- Adds support for recognizing GitLab.com repositories and correctly inferring `baseContentUrl` and `baseImagesUrl`. 
- Adds support for recognizing and replacing GitLab issue references with the correct URL

Caveats:

- GitLab allows subgroups, so `https://gitlab.com/foo/bar/baz.git` could be a valid repository.
- GitLab is adding a [special references](https://docs.gitlab.com/ee/user/markdown.html#special-gitlab-references) syntax, so in the future, issue references may also come in the form `[issue:123]`

I just duplicated the flags `--githubBranch` and `--no-gitHubIssueLinking` and associated code paths, `s/github/gitlab/`. It may be better to rename the flags instead of duplicating them, but I couldn't think of a good name for "Hosted Git server branch name". I thought about sharing `--githubBranch`, but IMO that would be bad UX (for `--githubBranch` to be relevant to GitLab repos).

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -82,29 +82,48 @@ module.exports = function (argv: string[]): void {
 			'--githubBranch [branch]',
 			'The GitHub branch used to infer relative links in README.md. Can be overriden by --baseContentUrl and --baseImagesUrl.'
 		)
+		.option(
+			'--gitlabBranch [branch]',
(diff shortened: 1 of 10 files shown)


joaomoreno (approved):
Thanks for the tests, they really helped to get this merged in without hassle. No worries about the duplication, I think I would've done the same. Thanks! 🍻

Issue #489, opened 2020-09-21
Repository shorthand


cbarrick:
With npm, `package.json` supports a shorthand for GitHub, GitLab, BitBucket, etc.

https://docs.npmjs.com/files/package.json#repository

However, vsce spits out a warning when one of these shorthands is used:

```shell
$ cat package.json | grep repository
    "repository": "github:foo/bar",

$ vsce package
 ERROR  Couldn't detect the repository where this extension is published. The link './LICENSE' will be broken in README.md. Please provide the repository URL in package.json or use the --baseContentUrl and --baseImagesUrl options.
```

It would be nice if vsce would support these shorthands.

wangweixuan:
I suggest including the package [`hosted-git-info`](https://github.com/npm/hosted-git-info), which supports parsing those shorthands and is developed by npm.

joaomoreno:
@wangweixuan Please close one of the PRs, pick your favorite.

PR #561, opened 2021-04-23
update regex to support repository shorthand

abhijit-hota:
potential fix for #489


joaomoreno:
This was just missing a test case. Otherwise, all good, thanks! 🍻

joaomoreno (approved):

abhijit-hota:
Thanks!
P.S. I wasn't aware of adding the test case so couldn't add it. 😅

PR #564, opened 2021-05-04
fix: Missing error response handling

### LINKED ISSUE #563: Cannot read property '0' of undefined
The very unhelpful error message `Cannot read property '0' of undefined` is printed when executing `vsce show foo`.

oliversalzburg:
In case of an invalid extension identifier, the error message from the backend is now properly displayed in the console:

> ERROR  The query supplied was not valid, one of the filters contained an error. The fully qualified name MUST contain a '.' between the publisher and extension names.

image

Fixes #563


joaomoreno:
Thanks! 🍻

PR #577, opened 2021-06-07
fix .vscodeignore not working

fyangstudio:
#576 

Update ignore usage:

```js
ignore()
.add(fs.readFileSync(filenameOfGitignore).toString())
.filter(filenames)
```

see: https://www.npmjs.com/package/ignore

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "vsce",
-	"version": "1.92.0",
+	"version": "1.92.1",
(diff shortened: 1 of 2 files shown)


joaomoreno:
This doesn't change anything at all besides using sync Node API.

fyangstudio:
ignore().add only work with:

fs.readFileSync(filenameOfGitignore).toString()

not work with your code.  please see #576

PR #602, opened 2021-09-01
validate extension kind

sandy081:
Validate and throw error if extension kind has invalid values

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -1032,6 +1032,16 @@ export function validateManifest(manifest: Manifest): Manifest {
 		}
 	});
 
+	if (manifest.extensionKind) {
+		const extensionKinds: ExtensionKind[] = Array.isArray(manifest.extensionKind)
+			? manifest.extensionKind
+			: [manifest.extensionKind];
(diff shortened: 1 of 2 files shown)


src/package.ts
```diff
@@ -1032,6 +1032,16 @@ export function validateManifest(manifest: Manifest): Manifest {
 		}
 	});
 
+	if (manifest.extensionKind) {
+		const extensionKinds: ExtensionKind[] = Array.isArray(manifest.extensionKind)
+			? manifest.extensionKind
+			: [manifest.extensionKind];
+		const validExtensionKinds: ExtensionKind[] = ['ui', 'workspace'];
+		if (extensionKinds.some(extensionKind => !validExtensionKinds.includes(extensionKind))) {
+			throw new Error(`Manifest has invalid values for 'extensionKind' property. Allowed values are 'ui', 'workspace'`);
```
joaomoreno, line 1041:
```suggestion
			throw new Error(`Manifest contains invalid value '${extensionKind}' in the 'extensionKind' property. Allowed values are 'ui', 'workspace'.`);
```

joaomoreno (approved):
Looks good, we should just print a better error indicating what's the value that throws the error.

Issue #605, opened 2021-09-03
Searching via vsce returns ADO and VSIDE extensions along with VS Code extensions


mayankmaheshwari18:
When we search from VSCE using `vsce search <search text>`, it returns all VSIDE, ADO extensions too.

Repro steps:
1. Open cmd prompt and enter vsce search test-extension
2. List returns extensions which are ADO and VSIDE extensions.
![image](https://user-images.githubusercontent.com/16611785/131979301-feae20cb-363e-4fde-8198-0ff2d8e0985d.png)

Highlighted Extensions:
1. ADO: https://marketplace.visualstudio.com/items?itemName=westpeaksconsulting.vsts-test-extensions
2. VSIDE: https://marketplace.visualstudio.com/items?itemName=mcferden.TestExtension

Expectations: VSCE should return only vs code extensions.

PR #606, opened 2021-09-03
Fixed an issue where search returns ADO and VSIDE

prashantvc:
Fixed the issue #605 where search returns ADO, VSIDE, un-published, or locked extensions along with VS Code extensions


joaomoreno:
Thanks!

Issue #625, opened 2021-09-30
VSCE should automate publishing the same bits for multiple platforms at once


joaomoreno:
From @isidorn

Based on feedback from @aeschli the `--target` should be more user friendly so extension publishers can more easily publish multiple platforms.
Thus I suggest that the `--target` also supports specifying an array, such that the user can list multiple platforms.

Other ideas from @aeschli 
Specify the list of platforms in my manifest.

PR #639, opened 2021-10-15
package: Introduce `list-targets` flag

radeksimko:
This is one way of addressing some problems outlined in https://github.com/microsoft/vscode-vsce/issues/625 but more importantly make it easier for end-users to see what targets are even supported without having to look into the source code.

I'm half wondering if this is better implemented as a new subcommand? `vsce ls-targets` or `vsce ls-package-targets`, I'm assuming that this would allow easier expansion with more flags later. I imagine that a `--json` flag might be useful here to change the shell-friendly new-line-separated output to JSON.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -79,6 +79,7 @@ module.exports = function (argv: string[]): void {
 		.description('Packages an extension')
 		.option('-o, --out <path>', 'Output .vsix extension file to <path> location (defaults to <name>-<version>.vsix)')
 		.option('-t, --target <target>', 'Target architecture')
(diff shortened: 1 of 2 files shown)


joaomoreno:
> So to clarify the use case - we wouldn't just simply call `vsce package --target=$TARGET` but also make sure that the right binary is bundled in that package relevant for that target.

We suggest to manually include the targeted platforms, as per our sample: https://github.com/microsoft/vscode-platform-specific-sample/blob/main/.github/workflows/ci.yml

Issue #665, opened 2021-12-14
--pre-release is ignored when publishing with --packagePath


roblourens:
The SSH extension is built and published using a vsix and the --packagePath flag, for example the publish command looks like `npx vsce publish --pre-release --packagePath $path --noVerify`. Apparently when publishing using that flag, the `--pre-release` flag is silently ignored and the extension is published as a stable version. Looking over the vsce code, I don't see the problem.

joaomoreno:
Fixed by https://github.com/microsoft/vscode-vsce/pull/666

PR #666, opened 2021-12-15
validate the package if pre-release flag is passed

sandy081:
Fixes - https://github.com/microsoft/vscode-vsce/issues/665

validate the package if pre-release flag is passed and the package is not a pre-release version

### DIFF
--- a/src/publish.ts
+++ b/src/publish.ts
@@ -51,6 +51,22 @@ export async function publish(options: IPublishOptions = {}): Promise<any> {
 				throw new Error(`Invalid extension VSIX manifest. ${err}`);
 			}
 
+			if (options.preRelease) {
+				let isPreReleasePackage = true;
+				try {
+					isPreReleasePackage = !!vsix.xmlManifest.PackageManifest.Metadata[0].Properties[0].Property.some(
+						p => p.$.Id === 'Microsoft.VisualStudio.Code.PreRelease'
+					);
+				} catch (err) {
+					throw new Error(`Invalid extension VSIX manifest. ${err}`);
+				}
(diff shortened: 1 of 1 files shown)


src/publish.ts
```diff
@@ -51,6 +51,22 @@ export async function publish(options: IPublishOptions = {}): Promise<any> {
...
+				try {
+					isPreReleasePackage = !!vsix.xmlManifest.PackageManifest.Metadata[0].Properties[0].Property.some(
+						p => p.$.Id === 'Microsoft.VisualStudio.Code.PreRelease'
+					);
+				} catch (err) {
+					throw new Error(`Invalid extension VSIX manifest. ${err}`);
+				}
+				if (!isPreReleasePackage) {
+					throw new Error(
+						`Cannot use '--pre-release' flag with the package that is not a pre-release. To publish the package as a pre-release version, please package it using '--pre-release' flag and try publishing again.`
```
joaomoreno, line 65:
```suggestion
						`Cannot use '--pre-release' flag with a package that was not packaged as pre-release. Please package it using the '--pre-release' flag and publish again.`
```

src/publish.ts
```diff
@@ -51,6 +51,22 @@ export async function publish(options: IPublishOptions = {}): Promise<any> {
 				throw new Error(`Invalid extension VSIX manifest. ${err}`);
 			}
 
+			if (options.preRelease) {
+				let isPreReleasePackage = true;
```
joaomoreno, line 55:
```suggestion
				let isPreReleasePackage = false;
```

joaomoreno (approved):

PR #704, opened 2022-03-07
docs: add configuration tips to readme

### LINKED ISSUE #703: Document configuration through `package.json`
Thanks to https://github.com/microsoft/vscode-vsce/commit/7182692b0f257dc10e7fc643269511549ca0c1db we can now configure `vsce` options directly in `package.json`.

It is, however, not documented. Some people are asking me about it (https://github.com/felipecrs/semantic-release-vsce/issues/260) and I lack a reference which I can share to them.

A very basic introduction in README or in the docs would be enough, I believe. No need to dive deep into it, but it would be nice to mention that the config keys are the same as the ones used in CLI, like `--baseUrl` (`package.json::vsce.baseUrl`).

felipecrs:
And also tidy up some other pieces.

Closes #703

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,13 +27,13 @@ Depending on your distribution, you will need to run the following command:
 
 Install vsce globally:
 
-```sh
-npm install -g vsce
+```console
+npm install --global vsce
 ```
 
 Verify the installation:
 
-```sh
+```console
 vsce --version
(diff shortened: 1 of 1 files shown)


joaomoreno:
Very cool thanks!

PR #720, opened 2022-04-14
feat: make command line more consistent and log bypass options in messages

wandyezj:
add `--no-verify` to be consistent with other no options.

log bypass option for:
--no-verify
--allow-star-activation
--allow-missing-repository


src/main.ts
```diff
@@ -183,6 +183,7 @@ module.exports = function (argv: string[]): void {
 		.option('--yarn', 'Use yarn instead of npm (default inferred from presence of yarn.lock or .yarnrc)')
 		.option('--no-yarn', 'Use npm instead of yarn (default inferred from lack of yarn.lock or .yarnrc)')
 		.option('--noVerify')
+		.option('--no-verify')
```
wandyezj, line 186: `--no-verify` is linked to using proposed APIs. Would it make sense to change this option to `--allow-proposed-apis` to match the other allow options?

src/main.ts
```diff
@@ -183,6 +183,7 @@ module.exports = function (argv: string[]): void {
 		.option('--yarn', 'Use yarn instead of npm (default inferred from presence of yarn.lock or .yarnrc)')
 		.option('--no-yarn', 'Use npm instead of yarn (default inferred from lack of yarn.lock or .yarnrc)')
 		.option('--noVerify')
```
wandyezj, line 185: There does not seem to be a convenient way to hide an existing option from help.

wandyezj, line 185: Can hide the old option with: https://github.com/tj/commander.js/issues/1716#issuecomment-1099615098

joaomoreno: Yup I suggest we do this.

src/main.ts
joaomoreno: Yeah I think that makes sense as well.

joaomoreno (changes requested):

joaomoreno (approved):

joaomoreno:
Thanks! 🍻

Issue #240, opened 2018-02-19
Trusted svg badge provider


equinusocio:
Hi, i noticed that **[opencollective.com](https://opencollective.com/)** is now a [trusted badge provider](https://github.com/Microsoft/vscode-vsce/blob/d478c021394c7be0d8ec2df3aae10c5c35799559/src/package.ts#L138) but i get this error when i publish my extension:

```
Error processing 'Equinusocio.vsc-material-theme' SVG reference in file '/extension/README.md'. Your reference to the SVG image ‘https://opencollective.com/vsc-material-theme/contributors.svg?width=890’ is not supported. Please see https://aka.ms/vsmarketplace_badge for more details.
```

Also opencollective.com doesn't appear here:
https://docs.microsoft.com/it-it/vsts/extend/develop/manifest#supported-badge-services

Using the vsce release.

joaomoreno:
Apparently, the Marketplace is also doing checks on the badges.

@viradhamMS @pkuma-msft Can we whitelist the following domains?

- https://github.com/Microsoft/vscode-vsce/commit/b3b198f0f7c7e65efb06c81d62e03afa76822960
- https://github.com/Microsoft/vscode-vsce/commit/75f7d2ffcb1706e59cc9796b1e36e298b55dbf4a
- https://github.com/Microsoft/vscode-vsce/commit/b25a7701456d7bcd8273e7e84ca9b656e7a1d74c

cc @mjbvz As far as I understood, these were the ones recently added.

pkuma-msft:
@equinusocio Sorry for the inconvenience, I'll add these domains to our whitelist and update this thread.

pkuma-msft:
We have made the change at our end, the publish should go through now. Please try and let us know if you still see any issues.

joaomoreno:
Thanks @pkuma-msft! 🍻

PR #722, opened 2022-04-24
feat: add readme path override

barnuri:
need the support of use different readme file when you pack\publish an extension 
i saw that there was an issue that close because it was out of scope, and i need it because my readme contains svgs, and it was very easy to add this support 
https://github.com/microsoft/vscode-vsce/issues/210

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -96,6 +96,7 @@ module.exports = function (argv: string[]): void {
 		.option('--no-rewrite-relative-links', 'Skip rewriting relative links.')
 		.option('--baseContentUrl <url>', 'Prepend all relative links in README.md with this url.')
 		.option('--baseImagesUrl <url>', 'Prepend all relative image links in README.md with this url.')
+		.option('--readmePath <path>', 'Set the path or readme file, by default use README.md')
 		.option('--yarn', 'Use yarn instead of npm (default inferred from presence of yarn.lock or .yarnrc)')
 		.option('--no-yarn', 'Use npm instead of yarn (default inferred from lack of yarn.lock or .yarnrc)')
 		.option('--ignoreFile <path>', 'Indicate alternative .vscodeignore')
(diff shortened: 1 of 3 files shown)


joaomoreno:
> and i need it because my readme contains svgs

Why exactly does the need for SVG warrant supporting different locations for README.md?

This is definitely out of scope, since changes would need to happen across the board: the Markteplace itself as well as VS Code, not just vsce.

barnuri:
> > and i need it because my readme contains svgs
> 
> Why exactly does the need for SVG warrant supporting different locations for README.md?
> 
> This is definitely out of scope, since changes would need to happen across the board: the Markteplace itself as well as VS Code, not just vsce.

because I have in the readme file a github action bagates, so i want to create the same readme without them and set him to the packing & deploying process in vsce

because vsce deploying failed on include svg in readme

barnuri:
so in this case you can manage a readme file for vscode and readme file for git

barnuri:
i find that there was an issue that talk about it too that you add some domain to whitelist, in my git i use github relative links of badges to make it work also it the forks of the main repo

https://github.com/Microsoft/vscode-vsce/issues/240

joaomoreno:
What domain do your svg icons come from?

barnuri:
> 

no like i said its a github actions bagdes, they give you an svg not an png
and i need them relative because i want them to work for any fork
why its a problem to add this feature to support different Readme file ?

[.github/workflows/runTests.yaml](/../../actions/workflows/runTests.yaml)
[.github/workflows/publish.yaml](/../../actions/workflows/publish.yaml)

joaomoreno:
It is incredibly error prone to use `..` in those paths. Just use the full URL to them and you'll be fine and VSCE won't complain.

> why its a problem to add this feature to support different Readme file ?

Because maybe it makes vsce happy, but VS Code won't know that your extension's README.md will be somewhere else. Also, the VS Marketplace won't be aware as well.

PR #725, opened 2022-05-02
build(release): improve release logic

felipecrs:
Changing the preset to `conventionalcommits` ensures that the commit message matches the current specification of Conventional Commits. Otherwise, the default is `angular` which is the AngularJS style. I always wanted that semantic-release shipped that by default, but unfortunately they don't have time to merge [my PR](https://github.com/semantic-release/semantic-release/pull/1836).

It also allows to use things like `fix!: drop support for something` to indicate that the commit is a breaking change.

I also configured semantic-release with some things that I judge to be more optimal, like triggering releases when a production dependency gets updated, or a refactoring or performance improvement gets merged. I found many versions released with non helpful changelogs because of this, like [v2.7.0](https://github.com/microsoft/vscode-vsce/releases/tag/v2.6.4), and [v2.6.4](https://github.com/microsoft/vscode-vsce/releases/tag/v2.7.0).

It also reduces the clutter from the changelog messages by skipping the update of development dependencies.

To finalize, the new configuration improves the GitHub Releases, since they now will get the npm `.tgz` files as assets and there will be a link to the npmjs page of the release (like [here](https://github.com/felipecrs/semantic-release-vsce/releases/tag/v5.0.11#:~:text=This%20release%20is%20also%20available%20on%3A)).

I hope you like it! :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,6 +51,7 @@
 				"@types/xml2js": "^0.4.4",
 				"@types/yauzl": "^2.9.2",
 				"@types/yazl": "^2.4.2",
+				"conventional-changelog-conventionalcommits": "^4.6.3",
 				"husky": "^7.0.4",
 				"mocha": "^9.2.0",
 				"npm-run-all": "^4.1.5",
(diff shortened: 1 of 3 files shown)


felipecrs:
Oh, and by the way, I'm using most of these configs in the VSCode ShellCheck extension repository (along with the semantic-release-vsce plugin). So, feel free to browse the releases section of it to see some in action:

https://github.com/vscode-shellcheck/vscode-shellcheck/releases

felipecrs:
Not applicable any longer since https://github.com/microsoft/vscode-vsce/pull/791.

PR #740, opened 2022-06-07
feat: support cjs entry `main: "./xxx.cjs"` in package.json

ShenHongFei:


ShenHongFei:
VS Code will load it.
I wrote a plugin that was packaged with webpack, and the index.cjs file was generated after packaging
The `type` in the package.json generated after the build is module, and the `main` is index.cjs . After publishing, it can be loaded by VS Code and used normally
https://github.com/dolphindb/vscode-extension

seethroughdev:
This should probably be reopened.  I was also needing to load a `*.cjs` file as `main`, and can report that it works as expected locally, as well as in @ShenHongFei 's fork.

Can you please consider re-opening and adding this change?

Thanks.

lucono:
The current version of [Karma Test Explorerer](https://github.com/lucono/karma-test-explorer) uses a [.cjs entrypoint](https://github.com/lucono/karma-test-explorer/blob/2106bb34d608e8bf38ba09a9d06a43d00e6b847e/package.json#L34) as well, and loads without any issues from what I can tell. See my comment [here](https://github.com/microsoft/vscode/issues/130367#issuecomment-1409716641).

ShenHongFei:
I don't have permission to reopen this PR. @joaomoreno

joaomoreno:
Hi @ShenHongFei, sorry for misunderstanding this in the first place. This PR got merged which does the same: https://github.com/microsoft/vscode-vsce/pull/825

Issue #742, opened 2022-06-17
Dependencies cannot be configured inside package.json


ssbarnea:
I think that when this bug was fixed the ability to load the dependencies from package.json file did not work.

Based on REAME.md file, we were supposed to be able to configure inside `package.json`:
```
{
  ...
  "vsce": {
    "dependencies": false,
    "useYarn": false
  }
}
```

While `useYarn` works as expected and allow you to avoid having to add `--no-yarn`, the `dependencies` option has not effect at all, preventing you from fixing the package command from inside config.

That is indeed a big deal if you do not call `vsce package` yourself and not having ability to add the required extra arguments.

I tried to look at the https://github.com/microsoft/vscode-vsce/commit/d797b8d44b1d2571b9aee9dc47032615a45b9c56 to spot the bug but I was not able to find exact what was missed.

Note: PR incoming, already fixed it locally.

PR #743, opened 2022-06-17
fix: ensure that we can define dependency argument in config

ssbarnea:
Commander documentation states that when using negative commands, if you do not already have a non-negative command it sets the default to `true`, which is what broke the logic for loading defaults from `package.json`.

Fixes: #742
Reference: https://www.npmjs.com/package/commander#other-option-types-negatable-boolean-and-booleanvalue

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ Or you can also set them in the `package.json`, so that you avoid having to rety
 {
   "vsce": {
     "baseImagesUrl": "https://my.custom/base/images/url"
+    "dependencies": true,
+    "yarn": false
   }
 }
 ```
--- a/src/main.ts
+++ b/src/main.ts
@@ -70,7 +70,9 @@ module.exports = function (argv: string[]): void {
 			undefined
 		)
 		.option('--ignoreFile <path>', 'Indicate alternative .vscodeignore')
-		.option('--no-dependencies', 'Disable dependency detection via npm or yarn')
+		// default must remain undefined for dependencies or we will fail to load defaults from package.json
+		.option('--dependencies', 'Enable dependency detection via npm or yarn', undefined)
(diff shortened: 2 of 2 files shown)


ssbarnea:
@joaomoreno If you can take a look at this it would be awesome, that is a real blocker for our switch from npm to yarn for building, as we cannot add custom command line arguments to vsce calls.

README.md
```diff
@@ -74,6 +74,8 @@ Or you can also set them in the `package.json`, so that you avoid having to rety
 {
   "vsce": {
     "baseImagesUrl": "https://my.custom/base/images/url"
+    "dependencies": true,
+    "yarn": false
```
joaomoreno, line 78: It is really necessary to change the docs for this?

joaomoreno (changes requested):

README.md
ssbarnea: I would that that is highly desirable. It took two people on our side to investigate/guess how arguments are translated to settings entries. In fact I would be happy to see full setting example with all supported settings.

Still, if you want this the docs change to be separated it would be fine, I can do it. I just explained why updating documented example is important. Also it would be ok to put default values there instead of custom ones, for example we could put "dependencies: null" which I guess is the correspondent of undefined.

Let me know in which direction to take it.

joaomoreno: All good and your reasoning makes sense, thanks!

joaomoreno (approved):

ssbarnea:
Sadly I am unable to request the failed job rerun. I am sure it was a glitch and that has nothing to do with this PR.

ssbarnea:
Please approve GHA run.

ssbarnea:
Any chance to get this merged? Asking as I do not want to miss it from the next release. Thanks!

Issue #748, opened 2022-06-27
Support free trial label for extensions in the Marketplace


prashantvc:
Allow extension authors to indicate extensions are `Free` or `Free Trial` in the Marketplace
Marketplace will read pricing value from the manifest property `Microsoft.VisualStudio.Services.Content.Pricing`

joaomoreno:
@prashantvc What values are possible for the `Microsoft.VisualStudio.Services.Content.Pricing` property? Is it exactly `Free` and `Free Trial` only?

joaomoreno:
Nevermind, just saw the PR.

prashantvc:
> @prashantvc What values are possible for the `Microsoft.VisualStudio.Services.Content.Pricing` property? Is it exactly `Free` and `Free Trial` only?

No, there is `Paid` as well (for VS and AzDO extensions), but we are limiting `Free` and `Free Trial` for VS Code

prashantvc:
@joaomoreno we may want to close this issue. It's been live on the Marketplace for a while

PR #749, opened 2022-06-27
feat: support pricing model in the manifest file

prashantvc:
Added the support for extension authors label an extension as `FREE` or `FREE TRIAL`

#748

### DIFF
--- a/src/manifest.ts
+++ b/src/manifest.ts
@@ -79,6 +79,7 @@ export interface Manifest {
 	dependencies?: { [name: string]: string };
 	devDependencies?: { [name: string]: string };
 	private?: boolean;
+	pricing?: string;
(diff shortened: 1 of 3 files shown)


joaomoreno:
@sandy081 Could you drive this? Notice the introduction of a new field in `package.json`, which might be more impactful than just vsce (VS Code, docs, etc). We can add it to the July plan if necessary.

sandy081:
Let's plan it for next milestone. It would be nice to also show this info in VS Code.

ProKashif (approved):
Nice man

prashantvc:
@joaomoreno could merge this change please

joaomoreno (approved):

PR #752, opened 2022-07-09
feat: github actions logging

nhedger:
This PR adds support for logging in GitHub Actions runners using annotations.

### Description

When the `GITHUB_ACTIONS` environment variable is set, instead of logging a message with `console.log`, a GitHub Actions workflow command is used to log the message.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
 			"version": "0.0.0",
 			"license": "MIT",
 			"dependencies": {
+				"@actions/core": "^1.9.0",
 				"azure-devops-node-api": "^11.0.1",
 				"chalk": "^2.4.2",
 				"cheerio": "^1.0.0-rc.9",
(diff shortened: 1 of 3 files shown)


joaomoreno (changes requested):
Cool idea! But I don't think it warrants bringing in additional dependencies. Luckily we can achieve the same by simply outputting in their expected format: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-notice-message

nhedger:
I actually went without the dependency in the first place but realized that it contains a function for escaping the data and thought that using it would keep it more consistent.

https://github.com/actions/toolkit/blob/c5278cdd088a8ed6a87dbd5c80d7c1ae03beb6e5/packages/core/src/command.ts#L80-L85

I'm happy to remove the dependency and import that single function if you still think importing the dependency is overkill.

joaomoreno:
Yeah, it's definitely better. It also seems that all we need escaping is the odd `%`.

joaomoreno (approved):

PR #757, opened 2022-07-19
fix: Adds support for Yarn 2

### LINKED ISSUE #517: Support yarn v2
In a repository using yarn 2, `vsce package` will fail trying to enumerate dependencies:

```shell
 ERROR  Command failed: yarn list --prod --json
```

To reproduce, in an extension repository, use:

```shell
yarn set version berry
yarn install
vsce package
```

randallb:
Closes #517

### DIFF
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -155,9 +155,17 @@ function selectYarnDependencies(deps: YarnDependency[], packagedDependencies: st
 }
 
 async function getYarnProductionDependencies(cwd: string, packagedDependencies?: string[]): Promise<YarnDependency[]> {
+	const versionString = await new Promise<string>((c, e) =>
+		cp.exec(
+			'yarn --version',
+			{ cwd, encoding: 'utf8', env: { ...process.env }, maxBuffer: 5000 * 1024 },
+			(err, stdout) => (err ? e(err) : c(stdout))
+		)
+	);
+	const yarnCommand = versionString[0] !== 1 ? "yarn info --recursive --dependents --json" : 'yarn list --prod --json'
 	const raw = await new Promise<string>((c, e) =>
 		cp.exec(
-			'yarn list --prod --json',
+			yarnCommand,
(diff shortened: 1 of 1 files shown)


randallb:
I think this should probably have tests but i don't know how to add them yet...

joaomoreno:
> @joaomoreno pinging you since you own #517... do you have test opinions?

Not specific ones... just that it should probably be tested. ;)

src/npm.ts
```diff
@@ -155,9 +155,17 @@ function selectYarnDependencies(deps: YarnDependency[], packagedDependencies: st
...
+			'yarn --version',
+			{ cwd, encoding: 'utf8', env: { ...process.env }, maxBuffer: 5000 * 1024 },
+			(err, stdout) => (err ? e(err) : c(stdout))
+		)
+	);
+	const yarnCommand = versionString[0] !== 1 ? "yarn info --recursive --dependents --json" : 'yarn list --prod --json'
 	const raw = await new Promise<string>((c, e) =>
 		cp.exec(
-			'yarn list --prod --json',
+			yarnCommand,
```
claui, line 168: Line 176 (where it says `Could not parse result of `[…]) should be updated with `yarnCommand` as well; otherwise the error message is going to be misleading.

claui, line 168: Are you sure that `yarn info --recursive --dependents --json` is a compatible drop-in replacement for `yarn list --prod --json`, anyway?

On Yarn v3.3.1, the former prints a list of JSON objects, for example:

```json
{"value":"yallist@npm:4.0.0","children":{"Version":"4.0.0","Dependents":["lru-cache@npm:6.0.0","minipass@npm:3.3.6","minipass@npm:4.0.0","minizlib@npm:2.1.2","tar@npm:6.1.13"]}}
```

At least on Yarn 3.3.1, none of those lines seem to contain anything like `"type":"tree"`. However, the code that follows seems to rely on just that. Consequently, `vsce package` now errors out for me in line 176, saying:

```plain
Could not parse result of `yarn list --json`
```

claui (changes requested):

PR #760, opened 2022-07-24
fix: package name when updating version

seraku24:
Instructions like "major"/"minor"/"patch" update the version number in
`package.json`. However, these instructions were being used for the
default version number instead of the updated value in the manifest.
Fixed by adding a conditional guard similar to that in
`getDefaultPackageName`.


joaomoreno (approved):

alexr00 (approved):

joaomoreno (approved):

joaomoreno:
Thanks! 🍻

PR #762, opened 2022-08-10
support symlinks

samdenty:
This PR adds supports for listing the files inside a symlinked extension.

Before, it would include files that were ignored. And include the realpath, vs the fake path

### DIFF
--- a/src/npm.ts
+++ b/src/npm.ts
@@ -225,7 +225,7 @@ export async function getDependencies(
 
 	try {
 		const link = await fs.promises.readlink(cwd);
-		const pathToReplace = path.resolve(cwd, link);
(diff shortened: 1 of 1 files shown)


joaomoreno (changes requested):
This just looks risky, given all the string manipulation of paths. Could we have some concrete tests for this?

samdenty:
> This just looks risky, given all the string manipulation of paths. Could we have some concrete tests for this?

How would I test this?

joaomoreno:
> This PR adds supports for listing the files inside a symlinked extension.

In the test suite, create a test which scaffolds a symlinked extension and then uses vsce to package it.

joaomoreno:
Closing due to being stale.

PR #788, opened 2022-11-24
feat: support for yarn 2 and yarn berry (3)

### LINKED ISSUE #517: Support yarn v2

Khaomi:
This PR is confirmed to work on yarn classic (1), yarn 2 and yarn berry (3).

However due to the limitation of yarn 2 and yarn berry (3)'s cli, it's not possible to filter out devdependencies.

I will mark this PR as draft until someone figure out how to do so, i have been searching for a way to filter it out for a while now and i am too afraid to do massive codebase change.

- [ ] filter out devDependencies

close #517

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -180,6 +180,6 @@ export function patchOptionsWithManifest(options: any, manifest: Manifest): void
 	}
 }
 
-export function ndjsonToJson(data: string): any {
+export function ndjsonToJson(data: string): string {
 	return `[${data.split('\n').filter(nonnull).join(',')}]`;
 }


xhayper:

joaomoreno:
> However due to the limitation of yarn 2 and yarn berry (3)'s cli, it's not possible to filter out devdependencies.

What is the limitation, exactly?

xhayper:
Currently, the CLI don't have the `list` command, so we move to using `yarn info` which list out dependencies BUT far more limited, we can't filter out devdependencies using `info` command BUT that's the only way to list all dependencies as of right now.

xhayper:
I found out that there's an API for the yarn package manager, i will use it to figure out if we can filter out devdependencies.

https://github.com/yarnpkg/berry/blob/master/packages/plugin-essentials/sources/commands/info.ts

xhayper:
i am having trouble using the API, i don't know how to correctly implement the feature using the yarn api at all, it's too confusing for me to understand.

Vap0r1ze:
I think you could filter them out manually by checking the `package.json` in `cwd`

PR #791, opened 2022-12-02
Adopt NPM build & publish pipeline

joaomoreno:
- Create Azure Pipeline
- Remove GH Action
- Rename to `@vscode/vsce`
- Drop support for semantic release (sorry @felipecrs, we have to comply with policy here. We had a good run!)


alexr00 (approved):

felipecrs:
That's alright! It makes sense. I'm happy that it helped for a while.

Thanks for mentioning me btw, good to be aware of the package renaming for semantic-release-vsce.

Issue #868, opened 2023-05-19
publish option --skip-duplicate doesn't consider target platforms


mbehr1:
If I want to release for multiple target platforms but use skipDuplicate to be able to rerun e.g. in case the marketplace had a timeout during release for some targets this fails after the first version due to:

https://github.com/microsoft/vscode-vsce/blob/c2e71d5bcee680b31d009cf17423d4fb64c1883c/src/publish.ts#L179

I think this line needs a similar and condition like ```sameVersion.some(v => v.targetPlatform === options.target)```.

Currently with skipDuplicate only one target platform will be released which is imho wrong.

drewbrokke:
@mbehr1 I'll give this one a shot. I ran into the same issue today.

drewbrokke:
@joaomoreno is there a test case that will also need updating with this one?

roblourens:
@joaomoreno should we build from source to verify or do you want to publish a beta?

joaomoreno:
Sorry about that, just published v2.20.0.

PR #873, opened 2023-06-21
(#868) publish option --skip-duplicate doesn't consider target platforms

drewbrokke:
Fixes https://github.com/microsoft/vscode-vsce/issues/868

### Problem:
When publishing multiple `vsix` packages with different target platforms using the `--packagePath` flag and the `--skip-duplicate` flag, only the first package is published. After that, all the others are considered duplicate versions and are not published.

### Solution:
Always consider the target platform along with the version when the `target` is present in the publish options

I did not see any relevant test cases for this code, so please let me know if I missed it and I will update the PR.

Thanks!

FYI @mbehr1

### DIFF
--- a/src/publish.ts
+++ b/src/publish.ts
@@ -173,17 +173,18 @@ async function _publish(packagePath: string, manifest: Manifest, options: IInter
 		}
 
 		if (extension && extension.versions) {
-			const sameVersion = extension.versions.filter(v => v.version === manifest.version);
+			const versionExists = extension.versions.some(v =>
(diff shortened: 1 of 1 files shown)


mbehr1 (commented):
thx! Lgtm. Didn't test it yet but code looks good/makes sense.

alexdima (approved):

drewbrokke:
Thanks all! 🥳

joaomoreno:
No... thank YOU @drewbrokke! 🍻

PR #878, opened 2023-07-06
docs: consolidated cli commands descriptions. Revised README

bandantonio:
1. Consolidated all the descriptions of cli commands (style + tense)
2. Revised README

Questions to clarify:
1. [`--noVerify`](https://github.com/microsoft/vscode-vsce/blob/main/src/main.ts#L192): has no description, and after checking on my extension, still have no clue on what exactly does this flag do.
2. [`--no-dependencies`](https://github.com/microsoft/vscode-vsce/blob/main/src/main.ts#L111): this flag has the default value as `undefined` on lines [77](https://github.com/microsoft/vscode-vsce/blob/main/src/main.ts#L77) and [196](https://github.com/microsoft/vscode-vsce/blob/main/src/main.ts#L196), but on line [111](https://github.com/microsoft/vscode-vsce/blob/main/src/main.ts#L111) the default value is not set (or missing). Should it be there?


bandantonio:
> This flag is meant to be used when publishing an extension which uses `enableProposedApi` or `enabledApiProposals` in its `package.json`. Ideally we'd rename this to `--allow-proposed-api`.

Thank you for clarifying this, I updated the description of the command.

However, there is still a bit of confusion about `noVerify`: according to the [using Proposed API](https://code.visualstudio.com/api/advanced-topics/using-proposed-api) docs, publishing extensions with proposed APIs is highly discouraged or even considered a bad practice:

> ... only available in Insiders distribution and cannot be used in published extensions

This phrase logically contradicts with the fact that `--noVerify` flag is available for the `publish` command and can be used to skip the restriction. And I have successfully published my test extension with `enabledApiProposals` in my `package.json`. Could you please elaborate on what are use cases for `--noVerify` on production?

Another question: Let's rename it to `--allow-proposed-api` in this PR or better create a consequent one?

joaomoreno:
> highly discouraged or even considered a bad practice

Not exactly. It's more like: we don't support it.

That being said, there are always exclusions to the rule. So we want to support this. Feel free to rename it to `--allow-proposed-api` in this PR.

joaomoreno:
I should've mentioned: we should still silently support `noVerify`. We don't want to break no one. 🥲

bandantonio:
@joaomoreno Thanks for mentioning this, I should've thought about it as well.
Just to clarify "silently support `noVerify`" - you mean listing only `--allow-proposed-api` and remove `noVerify` from the list or it's better to mention both options but add a note that `--allow-proposed-api` is the recommended one and `noVerify` is left only for backward compatibility?

joaomoreno:
We can remove it from the list, but still respect it if people use it.

joaomoreno (approved):

alexdima (approved):

joaomoreno:
Thanks! 🍻

PR #887, opened 2023-08-02
feat: add support for packaging and publishing without license file

### LINKED ISSUE #779: vsce package lacks --no-license option
Extensions can be published without a license but you have to manually confirm that you want to proceed without a license.

This is problematic in CI/CD pipelines.

vsce should offer a --no-license option for extensions that are published without a license

bandantonio:
Resolves #779

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -112,6 +112,7 @@ module.exports = function (argv: string[]): void {
 		.option('--pre-release', 'Mark this package as a pre-release')
 		.option('--allow-star-activation', 'Allow using * in activation events')
 		.option('--allow-missing-repository', 'Allow missing a repository URL in package.json')
+		.option('--skip-license', 'Allow packaging without license file')
 		.action(
 			(
 				version,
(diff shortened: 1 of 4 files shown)


kotborealis:
Just a suggestion: some cli-utilities have an `-y/--yes/--assume-yes` flag for usage in scripts, which automatically accepts all prompt.

image

I think it would be more intuitive to have this kind of flag.

kotborealis:
Looking closely, there's even some code already in place to auto-accept prompts.

[src/util.ts](https://github.com/microsoft/vscode-vsce/blob/main/src/util.ts):
```ts
export function read(prompt: string, options: _read.Options = {}): Promise<string> {
	if (process.env['VSCE_TESTS'] || !process.stdout.isTTY) {
		return Promise.resolve('y');
	}

	return __read({ prompt, ...options });
}
```

However, `VSCE_TESTS` is not documented (?) and I don't feel really confident to use it.

Maybe we could add something like `VSCE_YES`/`VSCE_ASSUME_YES` env variable?
```diff
-	if (process.env['VSCE_TESTS'] || !process.stdout.isTTY) {
+ 	if (process.env['VSCE_ASSUME_YES'] || !process.stdout.isTTY) {
```

joaomoreno (approved):

hediet (approved):

PR #912, opened 2023-11-21
Filter data from other platforms

### LINKED ISSUE #916: Support `--ignore-other-target-folders`
When using `vsce package` or `vsce publish` with `--target` it would be nice to have a `--ignore-other-target-folders` which automatically and recursively ignores all folders named the same as all other targets.

This allows extension authors to have:

```
package.json
linux-x64/file.txt
darwin-arm64/file.txt
web/file.txt
```

And with `vsce publish --target linux-x64 --ignore-other-target-folders`, only the following files would be included:

```
package.json
linux-x64/file.txt
```

whosafe:
Fixes #916

When the parameter target has a value, all files will be filtered based on the target value. All file directories are in the target, but directories with different values from the target will be excluded from this packaging. For example:
ui
 |-style
     |- linux-x64
     |      |- default.css
     |- win32-x64
            |- default.css

When using vsce package - t win32-x64 for packaging, the linux-x64 directory will not be packaged into the program。

当参数 target 有值时，会根据target值过滤所有文件，所有文件目录在Target中，但时和target值不相同的目录，将会再本次打包中排除 。 如果：
ui
 |-style
     |- linux-x64
     |      |- default.css
     |- win32-x64
            |- default.css

在使用vsce package -t win32-x64 进行打包时，linux-x64目录将不会打包到程序里


joaomoreno:
This is a bit too implicit.

Should we instead support multiple files, like `.vscodeignore.win32-ia32`?

whosafe:
We can use the multi-file filtering method, but this will require writing the names of the other nine non-included files into the .vscodeignore.win32-ia32 file, resulting in excessive repetitive additions. In fact, the target parameter is already used to differentiate the system for packaging and there is no need to perform additional filtering through a file.

我们可以使用多文件过滤的方式，但这样会需要在 .vscodeignore.win32-ia32 文件中将其他九种不打包的文件都写入，导致重复性的添加过多。实际上，target 参数就已经用于区分系统进行打包，无需再通过一个文件进行额外的过滤。

joaomoreno:
I've cleaned this up, added tests and put the behavior behind a `--ignore-other-target-folders` option, which must be used together with a `--target` option.

joaomoreno:
Fixes #916

whosafe:

whosafe (commented):
ok

joaomoreno (approved):

lszomoru (approved):

joaomoreno (approved):

lszomoru (approved):

PR #942, opened 2024-03-22
Use NodeApi

sbanni:
Change to use to the new azure-devops-node-api method for calling publisher signing endpoint.


sandy081 (approved):

sandy081:
Awesome @sbanni 🎉

PR #950, opened 2024-04-10
Retry on API request timeout

### LINKED ISSUE #926: Retry on API request timeout
## Background

Currently when the connection between where `vsce` is running and the Marketplace API times out, the publishing fails. This becomes especially problematic if packages for multiple platforms are being published at the same time, e.g.

```
vsce publish --pre-release --no-git-tag-version --packagePath $(find . -iname *.vsix)
```

(where `*.vsix` can match 8 files)

We ran into this recently in https://github.com/hashicorp/vscode-terraform/actions/runs/7540203435/job/20524327288#step:3:22

```
Publishing 'hashicorp.terraform (darwin-x64) v2.29.2024011611'...
Extension URL (might take a few minutes): https://marketplace.visualstudio.com/items?itemName=hashicorp.terraform
Hub URL: https://marketplace.visualstudio.com/manage/publishers/hashicorp/extensions/terraform/hub
Published hashicorp.terraform (darwin-x64) v2.29.2024011611.
Publishing 'hashicorp.terraform (web) v2.29.2024011611'...
Extension URL (might take a few minutes): https://marketplace.visualstudio.com/items?itemName=hashicorp.terraform
Hub URL: https://marketplace.visualstudio.com/manage/publishers/hashicorp/extensions/terraform/hub
Published hashicorp.terraform (web) v2.29.2024011611.
Publishing 'hashicorp.terraform (linux-arm64) v2.29.2024011611'...
Extension URL (might take a few minutes): https://marketplace.visualstudio.com/items?itemName=hashicorp.terraform
Hub URL: https://marketplace.visualstudio.com/manage/publishers/hashicorp/extensions/terraform/hub
Published hashicorp.terraform (linux-arm64) v2.29.2024011611.
Publishing 'hashicorp.terraform (linux-x64) v2.29.2024011611'...
Error: Request timeout: /_apis/gallery/publishers/hashicorp/extensions/terraform
Error: Process completed with exit code 1.
```

It leaves the extension in a half-published state, which is not ideal.

The easiest fix is to run this again with `--skip-duplicates` but it would be great if there was no need for such retry in the first place.

## Proposal

Have vsce retry on API failures like timeout automatically a few times with some backoff in between.

e.g. retry 3 times with ~5-10 seconds in between?

benibenj:
This pull request adds retry functionality to the publishing process in order to handle API request timeouts. Currently, when the connection between where `vsce` is running and the Marketplace API times out, the publishing fails. The proposed solution is to automatically retry the API request a few times with a backoff in between. 

For convinience, the `cockatiel` library was used which is why the node version had to be incresed to `>=16` 

fixes #926

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
 				"azure-devops-node-api": "^12.5.0",
 				"chalk": "^2.4.2",
 				"cheerio": "^1.0.0-rc.9",
+				"cockatiel": "^3.1.2",
 				"commander": "^6.2.1",
 				"form-data": "^4.0.0",
 				"glob": "^7.0.6",
@@ -774,6 +775,14 @@
 				"string-width": "^4.2.0",
 				"strip-ansi": "^6.0.0",
 				"wrap-ansi": "^7.0.0"
+			}
+		},
+		"node_modules/cockatiel": {
+			"version": "3.1.2",
+			"resolved": "https://registry.npmjs.org/cockatiel/-/cockatiel-3.1.2.tgz",
+			"integrity": "sha512-5yARKww0dWyWg2/3xZeXgoxjHLwpVqFptj9Zy7qioJ6+/L0ARM184sgMUrQDjxw7ePJWlGhV998mKhzrxT0/Kg==",
+			"engines": {
+				"node": ">=16"
 			}
 		},
 		"node_modules/code-point-at": {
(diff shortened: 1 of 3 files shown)


joaomoreno:
> which is why the node version had to be incresed to `>=16`

Remember to call this out with ⚠️ in the GH Release. It usually means you will have a few angry users that have to upgrade their CI. Node 14 is so old that it's unlikely, but still...

joaomoreno (approved):

PR #959, opened 2024-04-15
fix: use namespace import for `glob` package

aleclarson:
I was getting an error from the `promisify(glob)` call, as the `glob` import was being resolved to undefined at runtime. Switching to a namespace import like `import * as glob` and changing the reference to `glob.default` has fixed the issue for me.

_Tested on Node v19, v20, and v21_

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -6,7 +6,7 @@ import * as yazl from 'yazl';
 import { ExtensionKind, Manifest } from './manifest';
 import { ITranslations, patchNLS } from './nls';
 import * as util from './util';
-import glob from 'glob';
+import * as glob from 'glob';
 import minimatch from 'minimatch';
 import markdownit from 'markdown-it';
 import * as cheerio from 'cheerio';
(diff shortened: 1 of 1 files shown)


LLotme (approved):

Crystal-RainSlide:
And `glob`'s version is very outdated:
``` log
> npm i @vscode/vsce
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported

> npm -g ls inflight
`-- @vscode/vsce@2.29.0
  `-- glob@7.2.3
    `-- inflight@1.0.6
```

An update would be great.

Crystal-RainSlide (approved):
LGTM, but `glob`'s version is not supported, update it if possible.

LLotme (approved):

benibenj:
Fixed by updating dependency #1027

PR #986, opened 2024-06-03
enable signing in vsce using script

sandy081:
enable signing in vsce using script

- add `--sign` script argument for signing the manifest
- enable package and publish to sign the package using `@vscode/vsce-sign` package and sign script

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -117,7 +117,7 @@ module.exports = function (argv: string[]): void {
 		.option('--allow-star-activation', 'Allow using * in activation events')
 		.option('--allow-missing-repository', 'Allow missing a repository URL in package.json')
 		.option('--skip-license', 'Allow packaging without license file')
-		.option('--sign', 'Script to sign the VSIX package. VSIX manifest will be passed as an argument.')
+		.option('--sign', 'Script to sign the VSIX package. VSIX manifest will be passed as an argument. Generates <package-name>.signature.zip alongside the VSIX package.')
 		.action(
 			(
 				version,
--- a/src/main.ts
+++ b/src/main.ts
@@ -117,7 +117,7 @@ module.exports = function (argv: string[]): void {
 		.option('--allow-star-activation', 'Allow using * in activation events')
 		.option('--allow-missing-repository', 'Allow missing a repository URL in package.json')
 		.option('--skip-license', 'Allow packaging without license file')
-		.option('--signtool', 'Path to the VSIX signing tool. Will be invoked with two arguments: `SIGNTOOL <path/to/extension.signature.manifest> <path/to/extension.signature.p7s>`.')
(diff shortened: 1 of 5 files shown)


src/main.ts
```diff
@@ -117,6 +117,7 @@ module.exports = function (argv: string[]): void {
 		.option('--allow-star-activation', 'Allow using * in activation events')
 		.option('--allow-missing-repository', 'Allow missing a repository URL in package.json')
 		.option('--skip-license', 'Allow packaging without license file')
+		.option('--sign', 'Script to sign the VSIX package. VSIX manifest will be passed as an argument. Generates <package-name>.signature.zip alongside the VSIX package.')
```
joaomoreno, line 120: I suggest we change the language from "script" to "tool" and simplify the interface by passing two arguments to it:

```suggestion
		.option('--signtool', 'Path to the VSIX signing tool. Will be invoked with two arguments: `SIGNTOOL <path/to/extension.vsix> <path/to/output-signature.zip>`.')
```

src/package.ts
```diff
@@ -1840,6 +1843,26 @@ export async function pack(options: IPackageOptions = {}): Promise<IPackageResul
...
+			}
+			if (stderr) {
+				return e();
+			}
+			return c();
+		});
+		proc.stdout?.on('data', (data) => {
+			console.log(data.toString('utf8'));
+		});
+	});
```
joaomoreno, line 1861: This can be simplified by using `util.promisify` and `stdio: 'inherit'`. That way you can simply `await` and don't have to do any stream forwarding.

Docs: https://nodejs.org/api/child_process.html#child_processexecfilefile-args-options-callback

joaomoreno (changes requested):

src/package.ts
```diff
@@ -1840,6 +1843,20 @@ export async function pack(options: IPackageOptions = {}): Promise<IPackageResul
...
+export async function signPackage(packageFile: string, signScript: string): Promise<string> {
+	const packageFolder = path.dirname(packageFile);
+	const packageName = path.basename(packageFile, '.vsix');
+	const manifestFile = path.join(packageFolder, `${packageName}.signature.manifest`);
+	const signatureFile = path.join(packageFolder, `${packageName}.signature.p7s`);
+	const signatureZip = path.join(packageFolder, `${packageName}.signature.zip`);
+
+	await generateManifest(packageFile, manifestFile);
+	const { stdout } = await promisify(cp.execFile)(signScript, [manifestFile,  signatureFile]);
+	console.log(stdout);
```
joaomoreno, line 1855: Another idea, since using sync is more than fine here:

```suggestion
	await cp.spawnSync(signScript, [manifestFile,  signatureFile], { stdio: 'inherit' });
```

joaomoreno (approved):

PR #994, opened 2024-06-07
Support signing related features

### LINKED ISSUE #993: Support signing related features
Support following signing related features

- Generate the extension manifest from the extension vsix: `vsce generate-manifest --packagePath <package-path> --out <manifest-output-path>`
- Publish the extension with signed manifest: `vsce publish --packagePath <package-path> --manifestPath <manifest-path> --signaturePath <signature-path>`

sandy081:
fix #993

Support following signing related features

- Generate the extension manifest from the extension vsix: `vsce generate-manifest --packagePath <package-path> --out <manifest-output-path>`
- Publish the extension with signed manifest: `vsce publish --packagePath <package-path> --manifestPath <manifest-path> --signaturePath <signature-path>`

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,6 +1,6 @@
 import program from 'commander';
 import leven from 'leven';
-import { packageCommand, ls, Targets } from './package';
+import { packageCommand, ls, Targets, generateManifest } from './package';
 import { publish, unpublish } from './publish';
 import { show } from './show';
 import { search } from './search';
@@ -197,6 +197,8 @@ module.exports = function (argv: string[]): void {
 		)
 		.option('--no-update-package-json', 'Do not update `package.json`. Valid only when [version] is provided.')
 		.option('-i, --packagePath <paths...>', 'Publish the provided VSIX packages.')
+		.option('--manifestPath <paths...>', 'Manifest files to publish alongside the VSIX packages.')
+		.option('--signaturePath <paths...>', 'Signature files to publish alongside the VSIX packages.')
 		.option('--sigzipPath <paths...>', 'Signature archives to publish alongside the VSIX packages.')
 		.option('--sign-tool <path>', 'Path to the VSIX signing tool. Will be invoked with two arguments: `SIGNTOOL <path/to/extension.signature.manifest> <path/to/extension.signature.p7s>`. This will be ignored if --sigzipPath is provided.')
 		.option(
@@ -237,6 +239,8 @@ module.exports = function (argv: string[]): void {
 					gitTagVersion,
 					updatePackageJson,
 					packagePath,
+					manifestPath,
+					signaturePath,
 					sigzipPath,
 					githubBranch,
 					gitlabBranch,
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -298,6 +304,19 @@ module.exports = function (argv: string[]): void {
 		.option('-f, --force', 'Skip confirmation prompt when unpublishing an extension')
 		.action((id, { pat, azureCredential, force }) => main(unpublish({ id, pat, azureCredential, force })));
 
+	program
+		.command('generate-manifest')
+		.description('Generates the extension manifest from the provided VSIX package.')
+		.requiredOption('-i, --packagePath <path>', 'Path to the VSIX package')
+		.option('-o, --out <path>', 'Output the extension manifest to <path> location (defaults to .signature.manifest)')
```
joaomoreno, line 311: As far as I can understand, the fact that `.signature.manifest` is the default value is an implementation detail of vsce-sign. If we declare it here, I suggest to also enforce it here. I also think it's weird that `.signature.manifest` is the default name. Wouldn't `extension.manifest` be a better name, since it has nothing to do with signatures at this point?

```suggestion
		.option('-o, --out <path>', 'Output the extension manifest to <path> location (defaults to .signature.manifest)', 'extension.manifest')
```

src/publish.ts
```diff
@@ -118,12 +126,19 @@ export async function publish(options: IPublishOptions = {}): Promise<any> {
 
 			validateMarketplaceRequirements(vsix.manifest, options);
 
-			let sigzipPath = options.sigzipPath?.[index];
+			let sigzipPath: string | undefined;
+			if (options.manifestPath?.[index] && options.signaturePath?.[index]) {
+				sigzipPath = await createSignatureArchive(options.manifestPath[index], options.signaturePath[index])
```
joaomoreno, line 131:
```suggestion
				sigzipPath = await createSignatureArchive(options.manifestPath[index], options.signaturePath[index]);
```

src/publish.ts
```diff
@@ -118,12 +126,19 @@ export async function publish(options: IPublishOptions = {}): Promise<any> {
...
+			}
+
+			if (!sigzipPath) {
+				sigzipPath = options.sigzipPath?.[index];
+			}
```
joaomoreno, line 136: Shouldn't this condition be before the previous one? If `sigzipPath` is passed in options, do we really still need to call `createSignatureArchive`?

joaomoreno (changes requested):

src/main.ts
sandy081: I will call it like VSIX. `(defaults to <name>-<version>.vsix)`

src/main.ts
```diff
@@ -298,6 +304,19 @@ module.exports = function (argv: string[]): void {
 		.option('-f, --force', 'Skip confirmation prompt when unpublishing an extension')
 		.action((id, { pat, azureCredential, force }) => main(unpublish({ id, pat, azureCredential, force })));
 
+	program
+		.command('generate-manifest')
+		.description('Generates the extension manifest from the provided VSIX package.')
+		.requiredOption('-i, --packagePath <path>', 'Path to the VSIX package')
+		.option('-o, --out <path>', 'Output the extension manifest to <path> location ((defaults to <packagename>.manifest)')
```
joaomoreno, line 311:
```suggestion
		.option('-o, --out <path>', 'Output the extension manifest to <path> location (defaults to <packagename>.manifest)')
```

joaomoreno (approved):

PR #1027, opened 2024-07-19
Update deprecated dependencies

### LINKED ISSUE #1026: glob 7.* is going EOL
Warning when installing on node 20

```
warning @vscode/vsce > glob@7.2.3: Glob versions prior to v9 are no longer supported
```

benibenj:
This pull request updates the deprecated dependencies in the project. The `glob` dependency has been updated from version 7.0.6 to 11.0.0, and the `tmp` dependency has been updated from version 0.2.1 to 0.2.3. Additionally, the `@types/glob` dependency has been updated from version 7.1.1 to 8.1.0. These updates ensure that the project is using the latest versions of the dependencies and avoids any potential issues caused by deprecated functionality.

closes #959
closes #1026


lszomoru (approved):

xoriath:
hrm, the new glob requires node 20 | >= 22... so just rolling the minor version for this breaks a lot of things :)

mattkubej:
Similarly got bit by the `glob` upgrade. We're packaging a TypeScript server plugin with a VSCode Extension, which was installed via a symlink. From looking at their [changelog](https://github.com/isaacs/node-glob/blob/main/changelog.md), I think we likely encountered "exclude symbolic links to directories when follow and nodir are both set". So, package was not including our dependency.

I was able to workaround this by not installing with a symlink, but just wanted to share in case others hit this.

benibenj:
@xoriath thanks for mentioning this. You are right, we should actually create a major release for this. I'll create a new minor with the old glob version and publish a major with the node 20 requirement.

xoriath:
> @xoriath thanks for mentioning this. You are right, we should actually create a major release for this. I'll create a new minor with the old glob version and publish a major with the node 20 requirement.

Perfect, thanks a bunch! :)

PR #1040, opened 2024-08-22
Allow packaging extension without a publisher

### LINKED ISSUE #1029: Publisher required when only packaging an extension
https://github.com/microsoft/vscode-vsce/pull/1011 introduced a requirement that we should set a publisher even if we are just calling `vsce package` and have no intention to publish the artifact.
I wonder if it would be possible to reverse this behavior and allow packaging an extension without defining a publisher.

benibenj:
This pull request includes changes to the validation of the publisher field in the package manifest. Previously, the validation was triggered even if no publisher was set. With this change, the validation is only performed if a publisher is specified, allowing users to package an extension without a publisher for testing purposes.

fixes #1029

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -3,7 +3,7 @@ import * as path from 'path';
 import { promisify } from 'util';
 import * as cp from 'child_process';
 import * as yazl from 'yazl';
-import { ExtensionKind, Manifest } from './manifest';
+import { ExtensionKind, ManifestPackage } from './manifest';
 import { ITranslations, patchNLS } from './nls';
 import * as util from './util';
 import { glob } from 'glob';
@@ -57,7 +57,7 @@ export function read(file: IFile): Promise<string> {
 }
 
 export interface IPackage {
-	manifest: Manifest;
+	manifest: ManifestPackage;
 	packagePath: string;
 }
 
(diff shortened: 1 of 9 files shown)


aeschli (approved):

src/package.ts
```diff
@@ -1268,7 +1268,11 @@ export class ValidationProcessor extends BaseProcessor {
 
 export function validateManifest(manifest: Manifest): Manifest {
 	validateExtensionName(manifest.name);
-	validatePublisher(manifest.publisher);
+
+	// allow users to package an extension without a publisher for testing reasons
+	if (manifest.publisher) {
+		validatePublisher(manifest.publisher);
+	}
```
Tyriar, line 1275: Something seems off with the types here? If `manifest.publisher` does not have `| undefined` it should, if it does then `validatePublisher` should support validating an undefined publisher?

Tyriar (dismissed):

src/package.ts
benibenj: You are totally right, it's weird because we have different requirements for packaging and publishing but the types are only ensured for publishing currently. I'll create different types and validation for both requirement cases, possibly more debt work to come...

sandy081 (approved):

Issue #580, opened 2021-06-07
npm workspaces


panayot-cankov:
I've just finished moving several VSCode extensions into single monorepo using [npm v7 workspaces](https://docs.npmjs.com/cli/v7/using-npm/workspaces).

However when the `vsce` packages using `npm` (not `yarn`) it does internally:
```
npm list --production --depth=99999 --loglevel=error
```
And that command fails from the workspace to lookup common modules in the root.
[I logged an issue at npm](https://github.com/npm/cli/issues/3378)
Although I am not sure if this is `npm list` bug or feature, probably the `npm list` will have to be extended to look those modules and then maybe some additional work will be necessary in the `vsce` to package the files correctly.

Jolg42:
Got into this problem as well, the error was not obvious at all. See:
```
 ERROR  invalid relative path: extension/../../.eslintignore
npm ERR! Lifecycle script `package` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: package@31.0.3 
npm ERR!   at location: /home/runner/work/repo/repo/packages/vscode 
```

The error message here seems to be from the zipping library yazl, https://github.com/thejoshwolfe/yazl/blob/36b2a6a31a32637dca19b34e7b3295378e78a644/index.js#L368
 
The `npm list` call can be fount at https://github.com/microsoft/vscode-vsce/blob/main/src/npm.ts#L68

If there were more debug logs, it would have been easier to find out 😓

51ngul4r1ty:
Any progress on this?  I would also like to get VS Code Extensions working in a monorepo and it doesn't work for me.  I'm assuming that there's still work needed to fully support npm workspaces?

xt0rted:
I just ran into this after a few hours of cryptic error messages. I haven't done extensive testing yet but my workaround so far is to bundle with [`@vercel/ncc`](https://github.com/vercel/ncc) just like I do on my GitHub Actions and then package with `vsce package --no-dependencies`. Since ncc bundles your code and your dependencies into a single file the `--no-dependencies` flag seems to be a sufficient hack.

remcohaszing:
I ran into this issue too. But I figured out it's also possble to specify `vsce` options in `package.json`.

```jsonc
{
  // …
  "vsce": {
    "dependencies": false
  }
}
```

vritant24:
Just ran into this issue, it would great to have this documented.
@remcohaszing 's workaround fixed it for us

[...]

PR #1053, opened 2024-09-18
Fix regression with workdir symlinks

deribaucourt:
See my comment in https://github.com/microsoft/vscode-vsce/issues/580#issuecomment-2357922938
While upgrading from vsce 2.3 to 3.1, I had this regression which prevented building my package.
I use the following `package.json` extract:
```json
  "workspaces": [
    "src/lib"
  ]
```
And this folder is a symlink.

If a workdir was refering a symlink directory, we would pass it to yazl, which would error out because it's a directory.

glob recommends using `follow: true` to consider directory symlinks as directories in the collection.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -75,8 +75,9 @@ module.exports = function (argv: string[]): void {
 		.option('--dependencies', 'Enable dependency detection via npm or yarn', undefined)
 		.option('--no-dependencies', 'Disable dependency detection via npm or yarn', undefined)
 		.option('--readme-path <path>', 'Path to README file (defaults to README.md)')
-		.action(({ tree, yarn, packagedDependencies, ignoreFile, dependencies, readmePath }) =>
-			main(ls({ tree, useYarn: yarn, packagedDependencies, ignoreFile, dependencies, readmePath }))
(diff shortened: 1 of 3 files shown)


ije:
i also failed to upgrade to 3.1, and i'm using a local package:

```json
 "dependencies": {
    "typescript-esmsh-plugin": "./typescript-esmsh-plugin"
  }
```

deribaucourt:
Hello @ije, by using this PR you should be able to fix your issue. Replace your `package.json` vsce dependency with this commit:
```json
"@vscode/vsce":  "github:deribaucourt/vscode-vsce#3cd6ca0faaafd87b31d56f793b626b709ad36b26"
```
Once it is merged, you can use the next official releases.

benibenj:
Thanks for looking into this. I feel we need to be a bit careful with this which is why I would propose to have this behind a cli argument. Maybe `--follow-symlinks`?

deribaucourt:
Thanks for your review @benibenj 
I added a new configuration option for the commands `package`, `ls` and `publish`.
I did not test publish because I don't want to push right now, but the code is identical to the other commands.

benibenj (approved):

roblourens (approved):

Issue #1059, opened 2024-09-27
Publishing signed extension fails if file name contains special character


dtivel:
If either file name in a signed extension publish contains a special character, publish will fail with "Invalid part headers."  This is because the Content-Disposition header for each file part contains the original file name, and its filename attribute 

## Repro steps

1. `vsce publish --packagePath <FileNameWith@Character> --sigzipPath <FileNameWith@Character>`

## Expected result

Publish will succeed.

## Actual result

Publish fails with "Invalid part headers."

https://developer.mozilla.org/docs/Web/HTTP/Headers/Content-Disposition#as_a_response_header_for_the_main_body says:

> The quotes around the filename are optional, but are necessary if you use special characters in the filename, such as spaces.

CC @sandy081, @sbanni

connor4312:
@sandy081 are you able to verify this? Don't think we have any docs on how to sign extensions manually...

sandy081:
Yes, we cannot verify this locally. I will ask @dtivel to check if this bug is indeed fixed for those who are seeing this.

dtivel:
@sandy081, @connor4312, I verified locally using vsce 3.2.1 and a local VSM instance.

PR #1060, opened 2024-09-27
Quote `filename` value in `Content-Disposition` header

dtivel:
Resolve https://github.com/microsoft/vscode-vsce/issues/1059.

This change quotes the `filename` value of the `Content-Disposition` header to ensure that file names with special characters are handled properly.

CC @sandy081, @sbanni


sandy081 (approved):

sandy081:
@dtivel  Thanks for the fix

lszomoru (approved):

PR #1061, opened 2024-09-27
Added unpublish to api.ts

andrewlayer:
Exposed unpublish function to the `api.ts` file so it can be used as a node package.  We personally use vsce this way.

Notes:
- Exposes `unpublish` function and `IUnpublishOptions`
- Makes the choice to default `force` to true for IUnpublishOptions because no prompt is available when running as a node package.

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -108,19 +108,15 @@ export function publishVSIX(packagePath: string | string[], options: IPublishVSI
 }
 
 /**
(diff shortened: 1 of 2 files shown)


adrianstephens (approved):

benibenj (approved):
Looks good, thank you @andrewlayer

joaomoreno (approved):

PR #1067, opened 2024-10-12
fix for including excluded and vice versa

adrianstephens:
I reported this earlier today, but then thought I'd have a go fixing it, and... behold!
I won't be offended if you don't use this - I am not a javascript expert.

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -1688,8 +1688,8 @@ function collectFiles(
 							Promise.reject(err) :
 							// No .vscodeignore file exists
 							manifestFileIncludes ?
-								// include all files in manifestFileIncludes and ignore the rest
-								Promise.resolve(manifestFileIncludes.map(file => `!${file}`).concat(['**']).join('\n\r')) :
+								// ignore all, and then include all files in manifestFileIncludes
+								Promise.resolve(['**', ...manifestFileIncludes.map(file => file[0] === '!' ? file.slice(1) : `!${file}`)].join('\n\r')) :
 								// "files" property not used in package.json
 								Promise.resolve('')
 				)
@@ -1712,22 +1712,14 @@ function collectFiles(
 				// Combine with default ignore list
 				.then(ignore => [...defaultIgnore, ...ignore, ...notIgnored])
 
-				// Split into ignore and negate list
+				// Check the ignore list in order to filter out files
 				.then(ignore =>
-					ignore.reduce<[string[], string[]]>(
-						(r, e) => (!/^\s*!/.test(e) ? [[...r[0], e], r[1]] : [r[0], [...r[1], e]]),
-						[[], []]
-					)
-				)
-				.then(r => ({ ignore: r[0], negate: r[1] }))
+					files.filter(f=> ignore.reduce((keep, i) => i[0] === '!'
+						? keep || minimatch(f, i.slice(1), MinimatchOptions)
+						: keep && !minimatch(f, i, MinimatchOptions),
+						true
+					))
 
-				// Filter out files
-				.then(({ ignore, negate }) =>
-					files.filter(
-						f =>
-							!ignore.some(i => minimatch(f, i, MinimatchOptions)) ||
-							negate.some(i => minimatch(f, i.substr(1), MinimatchOptions))
-					)
 				)
 		);
 	});


adrianstephens:

benibenj:
Thank you for looking into this. We try to follow the behaviour of npm packages to some degree and it's currently unclear to me if npm supports exclude patterns in their `files` property. If they do, then we should add support for this. If they don't we need to make sure to think this through. It might still make sure to support this as we do not support combining "files" and ".vscodeignore" which npm does.

If we decide on taking this we will need to add more tests [here](https://github.com/andrewlayer/vscode-vsce/blob/38b75b5b515a04e857d9e772dccb97ac5c8af19b/src/test/package.test.ts#L194).

adrianstephens:
It was unclear to me too - all I saw in the documentation was that "File patterns follow a similar syntax to .gitignore, but reversed.", and unofficial sources like https://medium.com/trabe/control-what-you-publish-inside-your-npm-packages-e3ec911638b8

I assumed not supporting files and .vscodeignore together was a deliberate attempt to avoid confusion; and if they both have the same expressive power why do you need the .vscodeignore at all (other than for legacy reasons)?

When you say 'we will need to add more tests' is the 'we' me or you? :)

adrianstephens:
Just wanted to add, regardless of having exclude patterns in the files property, the real problem I was trying to fix was that I couldn't find a way to tell vsce that I wanted all the files in the 'assets' folder except the psds.

In .vscodeignore:
```
*
!dist/**/*.js
!assets/*
assets/*.psd
```
because the excludes are processed before the includes, it becomes
```
*
assets/*.psd
!dist/**/*.js
!assets/*
```
so the psd clause is overridden

AshtonStephens:
@benibenj I added a file to the unit tests with explicit exclusion in the `package.json` file.

```json
	"files": [
		"foo",
		"!**/exclude.me",  <-- I added
		"foo2/bar2/include.me",
		"*/bar3/**",
		"package.json",
		"LICENSE",
		"README.md"
	]
```

Then I added this file that's implicitly included by the `foo` path but is excluded based on the global filename match.

```
manifestFiles
├── LICENSE
├── README.md
├── foo
│   └── bar
│       ├── exclude.me <-- I added and is excluded
│       └── hello.txt
├── foo2
│   └── bar2
│       ├── ignore.me
│       └── include.me
├── foo3
│   └── bar3
│       └── hello.txt
├── logger.log
└── package.json
```

With these changes the tests all pass. Are there any other tests you'd like to see on this feature?

It definitely makes sense to be matching the `npm` behavior and I'm looking at their documentation and I see specifically the last bullet [here](https://github.com/npm/cli/wiki/Files-&-Ignores#details) states that doing exclusions with `!` with the files parameter is undefined.

> - "The consequences are undefined" if you try to negate any of the files entries (that is, "!foo.js"). Please don't. Use .npmignore.

I was going to say the documentation is inconsistent, but it actually consistently says this. AI tools, Gemini specifically, seem to very much disagree, so I think it's a matter of time before the `npm` team gets bug reports on this.

In the worst case we include this and we end up supporting something in vscode better than the actual tool, which is potentially misleading until the tool fixes it. On the other hand, it seems to me it's a matter of time until this is supported by npm, especially with Gemini being confused about it, and the ramifications of it becoming supported in npm and the vscode extension failing that is worse than the alternative. 

The docs say including an excludes with a `!` is undefined, and therefore exclusion falls into valid behavior when the behavior is undefined. So, future proofing the extension, I think it makes the most sense to consider merging it.

AshtonStephens (approved):
I reviewed and then added the test. Considering the test is very little and I did the reformatting I was going to request, I'm approving.

wileyj (approved):
lgtm

benibenj (commented):
I expanded the test to check if it works as expected. It does not. 

If I exclude a folder and include files in that folder, the file will be included which it should not.
It should first consider all include patterns (do not start with !) and then based on the files which are included, should exclude all files matching the exclude patterns (start with !)

adrianstephens:
> I expanded the test to check if it works as expected. It does not.
I don't know if there are definitive rules about how include/exclude rules work, especially in a situation where there are two lists working opposite to each other. My feeling was that the intuitive thing was to take the clauses in order, so I want A, but I don't want B, but I do want C, etc. It was the sorting of clauses that made 'everything in assets except psd's' impossible to express.

benibenj:
I strongly believe that the order in which the patterns are added to the `files` property should not impact the outcome. This is the same for `.gitIgnore`, `.npmIgnore` and `.vscodeIgnore` files. Making the outcome depending on the order, might make it a little more powerful, however, I fear that this would make it harder for users to understand and maintain. Defining your `files` should be a simple process.

adrianstephens:
Using .vscodeIgnore in order to make it work like an include list I started with 'ignore everything' and then built up what I wanted; but the sorting prevented that from working. This change was initially just to fix that; adding '!' support for package.json's 'files' was secondary.

The .gitignore manual says, 'An optional prefix "!" which negates the pattern; any matching file excluded by a previous pattern will become included again.'. I don't know how accurately that matches their implementation, but it certainly sounds like they're maintaining the order.

Sorting precludes .vscodeIgnore from expressing 'everything in assets except psds'; the 'files' list could still do it, but not 'everything in assets except psds, but I do want icon.psd'.

adrianstephens:
These are the results of some experimentation with .gitignore:
```
out/*
!out/api.js
```
keeps api.js
```
out/
!out/api.js
```
ignores api.js
```
!out/api.js
out/*
```
ignores api.js

Which tells me two things:
- First, that once a directory is ignored, it won't then consider anything inside it (for un-ignoring), *but* `out/*` does not ignore the directory itself, which allows previously ignored files to be un-ignored.
- Secondly, order is important. Ignoring `out/*` after un-ignoring `out/api.js` keeps api.js ignored.

If you agree with this interpretation, and you're happy for vsce to follow the same rules I can make the necessary modifications.
Note though that your added test:
```
"!fooExclude/**",
"fooExclude/includeNotPossible.txt",
```
*would* include includeNotPossible.txt under this scheme

PR #1078, opened 2024-11-09
Updated the semver comparison

### LINKED ISSUE #1077: Support Non-Semver Engine Versions in VSCE Validation and Comparison
The Regex used for checking valid engine version, can have values like 1.96.x, which is not a valid semver. 
When using these kinds of version, the semver version comparison fails. 

https://github.com/microsoft/vscode-vsce/blob/f1fc6d255d634849db9a7f28e59a031c49cd4918/src/validation.ts#L51-L53

https://github.com/microsoft/vscode-vsce/blob/f1fc6d255d634849db9a7f28e59a031c49cd4918/src/package.ts#L506

mohankumarelec:
Closes #1077

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -501,8 +501,13 @@ export class ManifestProcessor extends BaseProcessor {
 				throw new Error('Failed to parse semver of engines.vscode');
 			}
 
+			const minEngineVersion = semver.minVersion(engineSemver.range);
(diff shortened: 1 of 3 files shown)


mohankumarelec:
@benibenj I have added a test for checking version, please let me know any other further tests are required, Thanks

benibenj (approved):

PR #1100, opened 2024-12-05
Allow for reproducible .vsix packages

### LINKED ISSUE #906: Reproducible builds
Running the same build twice produces different archives as timestamps differ.

Please add an option `--reproducible` which would set the timestamp to a fixed value.

We are currently using the following workaround:
```
EXT_LOCAL_NAME=$(pwd)/my-extension.vsix
vsce package --allow-missing-repository -o ${EXT_LOCAL_NAME}

# Create reproducible archive by setting deterministic timestamp
TMPDIR=$(mktemp -d)
unzip -q ${EXT_LOCAL_NAME} -d ${TMPDIR}
rm -f ${EXT_LOCAL_NAME}
cd ${TMPDIR} && find -exec touch -d 1970-01-01T00:00:00Z "{}" \; && zip -q -r -oX ${EXT_LOCAL_NAME} *
rm -rf ${TMPDIR}
```

stevedlawrence:
Running the same build produces .vsix package that have the same content, but are not bit for bit the same, making it somewhat complicated to verify reproducible builds. Two changes are needed to fix this:

1. The mtime of each file added to the .vsix archive is included in each archive entry, so builds that happen at different times will have different entry timestamps. To fix this, if the SOURCE_DATE_EPOCH environment variable is defined, it it now used as entry timestamp value instead. Builds will now be reproducible as long as they set the same SOURCE_DATE_EPOCH value. If the environment variable is not defined or is not an integer, the current behavior is used.
2. The order that files are collected in preparation for packaging into the .vsix file is non-deterministic, which can lead to archives with the same content but in different orders. To fix this, files are sorted by archive entry name prior to adding.

Fixes #906

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -505,7 +505,7 @@ export class ManifestProcessor extends BaseProcessor {
 			if (!minEngineVersion) {
 				throw new Error('Failed to get minVersion of engines.vscode')
 			}
-			
+
 			if (target) {
 				if (engineSemver.version !== 'latest' && !semver.satisfies(minEngineVersion, '>=1.61', { includePrerelease: true })) {
 					throw new Error(
@@ -1796,7 +1796,7 @@ export function collect(manifest: ManifestPackage, options: IPackageOptions = {}
 	});
 }
 
-export function writeVsix(files: IFile[], packagePath: string): Promise<void> {
+function writeVsix(files: IFile[], packagePath: string): Promise<void> {
 	return fs.promises
 		.unlink(packagePath)
 		.catch(err => (err.code !== 'ENOENT' ? Promise.reject(err) : Promise.resolve(null)))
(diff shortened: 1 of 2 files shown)


stevedlawrence:
@microsoft-github-policy-service agree [company="Owl Cyber Defense"]

stevedlawrence:
@microsoft-github-policy-service agree company="Owl Cyber Defense"

stevedlawrence:
@microsoft-github-policy-service agree company="Owl Cyber Defense"

stevedlawrence:
> Could you add a test to make sure this works as expected. It will also make it easier for us to make sure we don't break this later on.

Done, I've also rebased this branch to ontop of the latest master.

joaomoreno (approved):
This is cool.

lszomoru (approved):

PR #1111, opened 2025-01-20
chore: update chalk

### LINKED ISSUE #1110: Feature request: Deno compatibility
Currently the compatibility with Deno is blocked by `chalk` dependency, which uses `__proto__` https://github.com/chalk/chalk/blob/v2.4.2/index.js#L154, which Deno disables by default.

> Deno made a conscious decision to not support `Object.prototype.__proto__` for security reasons.
https://docs.deno.com/runtime/reference/cli/unstable_flags/#--unstable-unsafe-proto

You can check that with the below:
```
$ deno -A npm:@vscode/vsce -h
error: Uncaught (in promise) TypeError: chalk_1.default.bgGreen.black is not a function
```

Deno optionally supports `--unstable-unsafe-proto` flag which allows the use of `__proto__`, and that makes `vsce` work in Deno:

```
$ deno -A --unstable-unsafe-proto npm:@vscode/vsce -h
Usage: vsce <command>
...
```

kt3k:
`chalk@2.4.2` uses `__proto__` and that makes the package incompatible with Deno.

This PR adds the Deno compatibility to vsce by updating chalk to v4 (Chalk removed the use of `__proto__` at v4 https://github.com/chalk/chalk/pull/387 )

You can check the fix by the following steps:

```
npm run build
deno -A --unstable-detect-cjs vsce
```

The above commands fail with the below error with the main:
```shellsession
$ deno -A --unstable-detect-cjs vsce
error: Uncaught (in promise) TypeError: chalk_1.default.bgGreen.black is not a function
```
It shows the correct Usage message with this fix.
```shellsession
$ deno -A --unstable-detect-cjs vsce
Usage: vsce <command>
...
```

fixes #1110

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
 				"@azure/identity": "^4.1.0",
 				"@vscode/vsce-sign": "^2.0.0",
 				"azure-devops-node-api": "^12.5.0",
-				"chalk": "^2.4.2",
+				"chalk": "^4.1.2",
 				"cheerio": "^1.0.0-rc.9",
 				"cockatiel": "^3.1.2",
 				"commander": "^12.1.0",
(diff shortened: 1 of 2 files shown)


kt3k:

PR #1130, opened 2025-04-21
Make the [Content_Types].xml file deterministic

stevedlawrence:
The order of the extension + mime types listed in the [Content_Types].xml file is dependent on the order files are listed in the files array, which can vary depending on the underlying filesystem's traversal order. This can lead to non-reproducible .vsix files.

This sorts the contentTypes array to ensure the order is always consistent and reproducible, regardless of the order of the files array, ensuring reproducible .vsix files.

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -1601,8 +1601,14 @@ export async function toContentTypes(files: IFile[]): Promise<string> {
 		contentTypes.push(`<Default Extension="${extension}" ContentType="${contentType}"/>`);
 	}
 
+	// The files array passed into this function has non-deterministic order
+	// depending on filesystem directory traversal. This leads to the order of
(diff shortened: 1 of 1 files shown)


benibenj (changes requested):
Could you add a comment inline mentioning why sort is being used.

stevedlawrence:
> Could you add a comment inline mentioning why sort is being used.

Done

lszomoru (approved):

stevedlawrence:
For planning purposes, is there an estimate on what release this will be part of and when?

benibenj:
We currently have an issue with community PR pipelines. Sorry about that. I'm hoping I'll have time to fix it until end of the week, but might be next week.

PR #1132, opened 2025-04-26
add --override-main-entrypoint

joshspicer:


meganrogge (dismissed):

src/package.ts
```diff
@@ -492,6 +497,10 @@ export class ManifestProcessor extends BaseProcessor {
 		const target = options.target;
 		const preRelease = options.preRelease;
 
+		if (options.overrideMainEntrypoint) {
+			manifest.main = options.overrideMainEntrypoint;
```
joaomoreno, line 501: `manifest` is supposed to be read-only here. Do we really need to write to it here?

src/main.ts
```diff
@@ -127,6 +127,7 @@ module.exports = function (argv: string[]): void {
 		.option('--skip-license', 'Allow packaging without license file')
 		.option('--sign-tool <path>', 'Path to the VSIX signing tool. Will be invoked with two arguments: `SIGNTOOL <path/to/extension.signature.manifest> <path/to/extension.signature.p7s>`.')
 		.option('--follow-symlinks', 'Recurse into symlinked directories instead of treating them as files')
+		.addOption(new Option("--override-main-entrypoint <path>", "Override the main entrypoint in the resulting packaged vsix").hideHelp(true))
```
joaomoreno, line 130: No reason not to document this.

src/package.ts
```diff
@@ -575,10 +584,17 @@ export class ManifestProcessor extends BaseProcessor {
...
 
-		if (this.options.version && !(this.options.updatePackageJson ?? true)) {
+		const needsUpdate = this.options.overrideMainEntrypoint || (this.options.version && !(this.options.updatePackageJson ?? true));
+
+		if (needsUpdate) {
 			const contents = await read(file);
 			const packageJson = JSON.parse(contents);
-			packageJson.version = this.options.version;
+			if (this.options.overrideMainEntrypoint) {
+					packageJson.main = this.options.overrideMainEntrypoint;
```
joaomoreno, line 593: Formatting.

src/package.ts
```diff
@@ -1081,18 +1097,18 @@ export class LicenseProcessor extends BaseProcessor {
...
 		}
 		if (manifest.browser) {
 			this.entryPoints.add(util.normalize(path.join('extension', this.appendJSExt(manifest.browser))));
 		}
 	}
 
-	appendJSExt(filePath: string): string {
-		if (filePath.endsWith('.js') || filePath.endsWith('.cjs')) {
+	appendJSExt(filePath: string, options: IPackageOptions = {}): string {
+		if (options.overrideMainEntrypoint || filePath.endsWith('.js') || filePath.endsWith('.cjs')) {
```
joaomoreno, line 1111: The difference in behavior should not be inside `appendJSExt`, but outside of it: if `overrideMainEntrypoint` is set, then `appendJSExt` should not be called.

src/test/fixtures/main/package.json
```diff
@@ -0,0 +1,7 @@
+{
+	"name": "main",
+	"publisher": "josh",
+	"version": "1.0.0",
+	"engines": { "vscode": "*" },
+    "main": "out/extension.js"
```
joaomoreno, line 6: Formatting.

src/test/package.test.ts
```diff
@@ -2309,6 +2309,26 @@ describe('ManifestProcessor', () => {
...
+		const processor = new ManifestProcessor(manifest, { overrideMainEntrypoint: 'foo.bar' });
+
+		const packageJson = {
+			path: 'extension/package.json',
+			localPath: path.join(root, 'package.json'),
+		};
+
+		// Update in-memory
+		manifest = JSON.parse(await read(await processor.onFile(packageJson)));
+		assert.deepStrictEqual(manifest.main, 'foo.bar');
```
joaomoreno, line 2326: What should be verified in test is that the `package.json` inside the VSIX should have the right value. Not that the in memory manifest that's thrown around in memory would contain the right value, since that's just an implementation detail.

joaomoreno (changes requested):

src/test/package.test.ts
joshspicer: done

src/package.ts
joshspicer: Ah yea, removed. This was covering an issue with my changes in `LaunchEntryPointProcessor`

joaomoreno (approved):
LGTM, would still love a review from @benibenj. He should be back on Friday.

PR #1152, opened 2025-05-21
Only scan secrets when .env is ignored and vice versa

benibenj:
```Copilot Generated Description:``` Adjust the scanning logic to only check for secrets when the corresponding options for packaging all secrets and .env files are not enabled. Update related error messages and linting configurations accordingly.


lszomoru (dismissed):

lszomoru:
@benibenj, there are conflicts 🔴

bpasero (approved):

Issue #1164, opened 2025-06-11
automatic tag generation conflict


theoephraim:
Some tags are automatically calculated during publishing, which means the limitations on tags (cannot start with special characters) bleeds into those properties, which I do not believe is desired or necessary.

In my case I am creating a new language contribution, and I'd like the human-readable label (set via the first `alias`) to start with an `@`. This works fine when installing locally, but publishing is blocked due to the automatic tag injection.

Several workarounds are possible but you could:
- allow the user to override the tags entirely
- allow the user to _remove_ an autogenerated tag
- automatically clean offending tags, rather than disallowing them, either by removing the tag or just removing the offending character

benibenj:
Thanks for reporting this. I will make sure we sanitize the alias before creating the tag. That means that the alias `@sql` will create the tag `sql`

rzhao271:
@theoephraim could you confirm the issue has been resolved for you with the latest vsce?

alexr00:
@benibenj is there a way to verify without publishing an extension?

PR #1169, opened 2025-06-16
Sanitize tags to remove special characters

benibenj:
Introduce a function to sanitize tags by removing non-word characters. Update the processing of language contributions to apply this sanitization to aliases, ensuring that tags conform to the required format.

Fixes microsoft/vscode-vsce#1164


joaomoreno (approved):

joaomoreno (approved):

theoephraim:
It appears this strips _all_ special characters, while the tag rule is actually just that it cannot _start_ with a special character. Also note that "." is allowed. Also might be worth considering why that's a restriction on tags at all?

Issue #1108, opened 2025-01-02
Feature Request: Add support for Bun package manager


coolcline:
**Is your feature request related to a problem? Please describe.**
Currently, vsce hardcodes the use of npm for running prepublish scripts (`npm run vscode:prepublish`). This causes issues when using alternative package managers like Bun, as the command fails with "npm: command not found" even when the project is properly configured to use Bun.

**Describe the solution you'd like**
I propose adding support for different package managers, specifically Bun, in the following ways:

1. Auto-detect the package manager being used in the project (by checking for bun.lockb, package-lock.json, yarn.lock, etc.)
2. Add a configuration option to explicitly specify which package manager to use (e.g., --package-manager=bun)
3. Use the detected/specified package manager when running scripts instead of hardcoding npm

**Describe alternatives you've considered**
Current workarounds include:
- Installing npm alongside Bun (not ideal, adds unnecessary dependencies)
- Using the --no-dependencies flag (doesn't work as the npm command is still hardcoded)

**Additional context**
Bun is gaining popularity as a fast, all-in-one JavaScript runtime and package manager. Supporting it would improve the developer experience for teams using Bun in their VS Code extension projects.

Example of a project using Bun:
```json
json
{
    "scripts": {
    "vscode:prepublish": "bun run package",
    "package": "bun run build:webview && bun run check-types && bun run lint"
    }
}
```

**Environment:**
- vsce version: 2.15.0
- OS: macOS
- Package Manager: Bun 1.1.42

Sorrow-Scarlet:
> [@Foorack](https://github.com/Foorack) running it with --no-dependencies also works.
> 
> `bunx vsce package --no-dependencies`

Thanks, confirmed working when compiling vscode-c3 extension.

asispts:
@VladyslavKochetkov it doesn't work

<img width="488" height="144" alt="Image" src="https://github.com/user-attachments/assets/f8302045-7990-4b46-b628-468cce18c73e" />

PR #1179, opened 2025-07-03
add `commonjs` annotation to package.json

Adjective-Object:
I'm working on making this tool compatible with [`deno`](https://deno.com/) and [`bun`](https://bun.sh/) for a personal project. It looks like I'm not the only one who would like to add that support (See: #1108)

Modern JS standards suggest either using a `.cjs` extension or a module `type` declaration on a package to know whether something is a commonjs module or an esm module. See: https://nodejs.org/api/packages.html#packagejson-and-file-extensions

This change adds the missing `"type": "commonjs"` annotation to the package.json file to allow `deno` to run the files correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
 		"type": "git",
 		"url": "https://github.com/Microsoft/vsce"
 	},
+	"type": "commonjs",
 	"homepage": "https://code.visualstudio.com",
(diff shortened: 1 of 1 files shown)


Adjective-Object:
@benibenj, do you know what the stalled `vsce OSS` check is? Not sure if I missed a setup step before I can make contributions

PR #1180, opened 2025-07-07
zip: gracefully handle failing to write a symlink

Adjective-Object:
yazl does not support writing symlinks into zipfiles. This causes issues when a user forgets to use `--follow-symlinks` while trying to write node_modules directories from other package managers into a vscode archive (e.g. when using `bun` or `deno`)

This change catches the error from yazl and improves the error message.

Before:
```
 ERROR  not a file: /workspaces/casty/node_modules/.deno/tree-sitter@0.19.0/node_modules/prebuild-install
```

After:
```
 ERROR  Tried to write a symlinked file (/workspaces/casty/node_modules/.deno/tree-sitter@0.19.0/node_modules/prebuild-install) to a zip archive. Did you forget to use '--follow-symlinks'?
```

### DIFF
--- a/src/package.ts
+++ b/src/package.ts
@@ -1846,8 +1846,28 @@ function writeVsix(files: IFile[], packagePath: string): Promise<void> {
 					const zipStream = fs.createWriteStream(packagePath);
 					zip.outputStream.pipe(zipStream);
 
-					zip.outputStream.once('error', e);
-					zipStream.once('error', e);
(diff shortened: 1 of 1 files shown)


benibenj:
I like that this would help the user by providing a better error message. I'm not convinced with the actual implementation. Also, Iet's make sure the error message from yazl is still shown.

PR #1181, opened 2025-07-07
main: expose `--packagedDependencies` to commands other than `ls`

Adjective-Object:
Currently, there's no way to expose `--packagedDependencies` when packaging from the CLI.
The CLI is only capable of querying the list of files with `vsce ls --packageDependencies`

This change adds the same CLI flag from `ls` to `package` and `publish`.

The user-facing benefit of this change is to allow publishing a limited subtree of dependencies from a javascript monorepo. Currently, a vscode extension in a javascript monorepo will publish the entire workspace's unfiltered dependency tree, as it has no way to specify `--packagedDependencies` during `publish` or `package`

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -114,6 +114,12 @@ module.exports = function (argv: string[]): void {
 		.option('--baseImagesUrl <url>', 'Prepend all relative image links in README.md with the specified URL.')
 		.option('--yarn', 'Use yarn instead of npm (default inferred from presence of yarn.lock or .yarnrc)')
 		.option('--no-yarn', 'Use npm instead of yarn (default inferred from absence of yarn.lock or .yarnrc)')
+		.option<string[]>(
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -235,6 +243,13 @@ module.exports = function (argv: string[]): void {
 		.option('--baseImagesUrl <url>', 'Prepend all relative image links in README.md with the specified URL.')
 		.option('--yarn', 'Use yarn instead of npm (default inferred from presence of yarn.lock or .yarnrc)')
 		.option('--no-yarn', 'Use npm instead of yarn (default inferred from absence of yarn.lock or .yarnrc)')
+		.option('--deno', 'Use deno instead of npm (default inferred from presence of deno.lock)')
```
benibenj, line 246: deno flag should not be part of this PR

PR #1182, opened 2025-07-07
package: collect dependencies from parent dirs in a monorepo

Adjective-Object:
The current packaging VSIX packaging strategy assumes that dependencies are set up in a standalone directory. However, if a vscode package were to exist in a `yarn` monorepo with [`nodeLinker: node-modules`](https://yarnpkg.com/features/linkers#nodelinker-node-modules) (e.g. one of the various large internal monorepos at Microsoft), it would not be able to package the extension.

This change adds support to that `yarn` wrapper for pulling dependencies from parent node_modules directories, if run inside of a yarn workspace.

It does this by replacing paths to node_modules folders in relative parent directories (for example `../../../node_modules/lodash/...`) into the extension's local node_modules directory in the archive (`node_modules/lodash/..`).


benibenj:
This seems reasonable and were fine with doing this. Will need to have a closer look at the changes when back from vacation

PR #1199, opened 2025-09-25
fix: generate language model tag for `languageModelChatProvider` contributions

joyceerhl:
Autogenerating `language-model` tags doesn't currently work, which seems to be because the contribution was renamed 4 weeks ago: https://github.com/microsoft/vscode/blob/f9618ba2a9749ee72df385329c56208eefdb3ef9/src/vs/workbench/contrib/chat/common/languageModels.ts#L290-L291

cc @benibenj @isidorn


isidorn (dismissed):
Thank you Joyce!

alexr00 (approved):

benibenj:
starting new vsce release

PR #1202, opened 2025-10-02
Replace keytar with @napi-rs/keyring/keytar

rtsao:
[@napi-rs/keyring](https://github.com/Brooooooklyn/keyring-node) exposes an [entry point](https://github.com/Brooooooklyn/keyring-node/blob/main/keytar.js) that is API compatible with `keytar`.

`@napi-rs/keyring` is preferable to `keytar` because it uses NAPI so prebuilt binaries work for any Node.js version. This plus the lack of `libsecret` dependencies makes this package much easier to install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,14 +15,8 @@ Read the [**Documentation**](https://code.visualstudio.com/api/working-with-exte
 
 ### Linux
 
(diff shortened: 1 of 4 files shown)


BenjaminBrienen:
This fixes a warning upon install about deprecated subdependencies found: `prebuild-install@7.1.3`. Please consider merging this.

jakebailey:
This can be closed, https://github.com/microsoft/vscode-vsce/pull/1305 did a similar thing.

PR #1205, opened 2025-10-04
Expose `verifyPat` to API

felipecrs:
Also as part of adoption in semantic-release-vsce, I noticed this API which I need was not exposed.

In case some outsider reads this PR, note that these APIs normally do not honor the environment variables like the CLI does. I.e. `VSCE_PAT` will not be honored by `verifyPat()` like `vsce verify-pat` does.


rickgonzalez:
An explainer scan using a tool I'm building for this PR - I hope it helps communicate the explanation. https://www.lenzon.ai/viewer/cmrjxqbpt0005cda8at7p1occ?voice=google-chirp3

PR #1270, opened 2026-04-29
update @azure/identity and @azure/msal-node 

BenjaminBrienen:
to remove transitive dependency on vulnerable uuid package

Fixes GHSA-w5hq-g745-h8pq


BenjaminBrienen:
Ok then

PR #1305, opened 2026-08-21
Update and replace many dependencies

jakebailey:
I'm interested this since vsce is now in the TypeScript dev dep tree, and has significant dep count and audit alerts.

Each commit is its own thing, which I can split into multiple PRs if desired. It's also technically a dupe of #1202, so it's ok if that happens and I rebase.

I'll also note that while I'm raising the minimum node engine to 20.19, the `@azure/identity` package did a bump to 22+ in a minor version, which means that vsce is technically also 22+ if someone relocks. So, there's a bunch more that could probably get updated.

I also have on another branch something that converts the repo to ESM, which is possible thanks to require(ESM). This does simplify some dep integration, but, is probably not worth much.

There are other deps I'd love to be able to replace with something more lightweight, like `glob`, but that's likely to cause breaking changes. Happy to discuss that, though.


src/publish.ts
```diff
@@ -1,22 +1,29 @@
...
 import FormData from 'form-data';
-import { basename } from 'path';
+import { basename, join } from 'path';
+import { tmpdir } from 'os';
 import { IterableBackoff, handleWhen, retry } from 'cockatiel';
 import { getAzureCredentialAccessToken } from './auth';
 import { getOIDCCredential } from './oidc';
 
-const tmpName = promisify(tmp.tmpName);
+async function withTemporaryPackage<T>(fn: (packagePath: string) => Promise<T>): Promise<T> {
```
jakebailey, line 18: This would be better as a `using` disposable, but that requires newer Node.

benibenj (approved):

benvillalobos (approved):

alexdima (approved):

PR #1313, opened 2026-08-28
Reduce secretlint dependency tree

jakebailey:
Much of the secretlint tree (44 deps) can be removed by just doing a little bit in vsce itself, as vsce doesn't actually use all of the config stuff, formatting, etc.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,9 +11,11 @@
 			"dependencies": {
 				"@azure/identity": "^4.1.0",
 				"@napi-rs/keyring": "^1.3.0",
-				"@secretlint/node": "^10.1.2",
+				"@secretlint/core": "^10.1.2",
 				"@secretlint/secretlint-rule-no-dotenv": "^10.1.2",
 				"@secretlint/secretlint-rule-preset-recommend": "^10.1.2",
+				"@secretlint/source-creator": "^10.1.2",
+				"@secretlint/types": "^10.1.2",
 				"@vscode/vsce-sign": "^2.0.0",
 				"azure-devops-node-api": "^12.5.0",
 				"chalk": "^4.1.2",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -1857,12 +1659,14 @@
...
 			"resolved": "https://registry.npmjs.org/fast-deep-equal/-/fast-deep-equal-3.1.3.tgz",
-			"integrity": "sha512-f3qQ9oQy9j2AhBe/H9VC91wLmKBCCU/gDOnKNAYG5hswO7BLKj09Hc5HYNz9cGI++xlpDCIgDaitVs03ATR84Q=="
+			"integrity": "sha512-f3qQ9oQy9j2AhBe/H9VC91wLmKBCCU/gDOnKNAYG5hswO7BLKj09Hc5HYNz9cGI++xlpDCIgDaitVs03ATR84Q==",
+			"dev": true
 		},
 		"node_modules/fast-uri": {
 			"version": "3.1.5",
 			"resolved": "https://registry.npmjs.org/fast-uri/-/fast-uri-3.1.5.tgz",
 			"integrity": "sha512-gHwA1O9LDIcKunMKhObS/HimwtehO1nPUECKAu5TpKgaO19fcWEl4bliWe1jWxVFvIXztJjjQ4L8XQ1EU9f7Jw==",
+			"dev": true,
```
jakebailey, line 1669: Doing this PR would stop everyone using `vsce` from having `fast-uri` related CVEs.

src/secretLint.ts
```diff
@@ -84,46 +63,75 @@ const dotEnvRules = [
...
+	let nextIndex = 0;
+
+	async function worker() {
+		while (nextIndex < values.length) {
+			const index = nextIndex++;
+			results[index] = await mapper(values[index]);
+		}
+	}
+
+	const workerCount = Math.min(os.cpus().length, values.length);
```
benibenj: AI Review: `os.cpus()` is documented to return `[]` when CPU information is unavailable. For a non-empty file list, that creates zero workers, so the sparse result array is treated as `{ ok: true, results: [] }` and packaging proceeds without scanning any file. The old `p-map` path rejected concurrency 0, so this changes a failure into a silent bypass. Please use `os.availableParallelism()` or otherwise guarantee at least one worker for non-empty input.

benibenj: Fixed in 7534898 by using `os.availableParallelism()` and adding a regression test for empty `os.cpus()`.

benibenj (approved):

lszomoru (approved):

src/secretLint.ts
jakebailey: I've never seen that happen before but availableParallelism is definitely better (it's just newer and I forgot about it 😄), thanks!

PR #1317, opened 2026-09-02
cli: add option to suppress new version warning

### LINKED ISSUE #1201: Please add `--no-new-version-warning` option
There is a warning whenever we run `vsce` because `vsce` is updated so often. An option like `--no-new-version-warning` would be helpful to suppress that warning, especially for CI systems.

Image

benibenj:
## Summary

- add a global `--no-new-version-warning` option
- skip the npm version check when the option is used
- preserve the existing warning behavior by default for every command

Fixes #1201

## Validation

- `npm run compile`
- `npm test` (249 passing, 3 pending)
- verified the option appears in help and is accepted before or after a subcommand

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -35,12 +35,12 @@ See https://code.visualstudio.com/api/working-with-extensions/publishing-extensi
 	process.exit(1);
 }
 
-function main(task: Promise<any>): void {
+function main(task: Promise<any>, showNewVersionWarning: boolean): void {
 	let latestVersion: string | null = null;
 
 	const token = new CancellationToken();
 
-	if (isatty(1)) {
+	if (showNewVersionWarning && isatty(1)) {
 		getLatestVersion(pkg.name, token)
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -61,8 +61,12 @@ const ValidTargets = [...Targets].join(', ');
 
 module.exports = function (argv: string[]): void {
 	const program = new Command();
+	const run = (task: Promise<any>): void => main(task, program.opts().newVersionWarning);
 
-	program.version(pkg.version).usage('<command>');
+	program
+		.version(pkg.version)
+		.usage('<command>')
+		.option('--no-new-version-warning', 'Do not show a warning when a newer version is available');
```
benibenj: make the comment a bit nicer

benibenj: Updated the help text to be clearer and more specific: “Suppress the warning shown when a newer version of vsce is available.”

> [!NOTE]
> Automated reply by VS Code Agent Merge.

PR #1324, opened 2026-09-03
Replace form-data with Node streams

jakebailey:
This can be replaced with a bit of code in the repo, deleting 8 prod deps.

I though it would be more but no, headers were already being manually constructed, so it's all just boilerplate with Node's streaming. There's more "handle errors in streams" code than actual form code, and then way more testing than actual form and error code.

### DIFF
--- a/src/multipart.ts
+++ b/src/multipart.ts
@@ -13,6 +13,11 @@ function escapeHeaderParameter(value: string): string {
 export function createMultipartStream(parts: readonly MultipartPart[], boundary: string): Readable {
 	const lineBreak = '\r\n';
 
+	for (const part of parts) {
+		// Readable's async iterator will rethrow stored errors when this part is consumed.
+		part.stream.on('error', () => {});
+	}
+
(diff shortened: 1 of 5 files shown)


src/multipart.ts
```diff
@@ -0,0 +1,56 @@
...
+		(async function* () {
+			for (const part of parts) {
+				yield `--${boundary}${lineBreak}`;
+				yield `Content-Disposition: attachment; name=${escapeHeaderParameter(part.name)}; filename="${escapeHeaderParameter(part.filename)}"${lineBreak}`;
+				yield `Content-Type: application/octet-stream${lineBreak}${lineBreak}`;
+				yield* part.stream;
+				yield lineBreak;
+			}
+
+			yield `--${boundary}--${lineBreak}`;
```
jakebailey: It still boggles my mind that this is how multipart form-data works

benibenj (approved):

Giuspepe (approved):

PR #1325, opened 2026-09-03
Remove leven dependency

jakebailey:
Donating my helper from `hereby`: https://github.com/jakebailey/hereby/blob/79e88d2d616eb199226d61f77e4c54bc8a40a1c1/src/cli/utils.ts#L61


benibenj (dismissed):

chrmarti (dismissed):
What's the advantage of having a copy of the helper instead of the maintained package?

benibenj (approved):

Giuspepe (approved):

